PR #291, opened 2023-07-25
Add URL Slug Validation

ann4bel:
For both entity and static pages, this pull request creates a function to check whether getPath() returns a valid or invalid (undefined or containing unsafe characters) URL slug value. If it's invalid, the page settings modal does not allow editing of the URL slug. If it's valid, then the modal does allow editing. However, if the user tries to save an invalid URL slug in the modal, then an error message is displayed and the Save button is disabled until new user input. 

For an entity page, a valid URL slug also includes any string that contains ${document.field} where "field" is variable. 

J=SLAP-2826
TEST=auto, manual

Checked that URL behavior works as expected in the test site.


packages/studio/src/components/AddPageButton/BasicPageDataCollector.tsx
```diff
@@ -41,6 +42,11 @@ export default function BasicPageDataCollector({
         const getPathValue = data.url
           ? createGetPathVal(data.url, isEntityPage)
           : undefined;
+        const pageDataValidator = new PageDataValidator();
```
tmeyer2115, line 45: Let's move the creation of `pageDataValidator` outside of the `onConfirm` callback. Each time `onConfirm` fires, we're creating a new instance of this class which is a bit inefficient. We can just instantiate it once, on creation of `BasicPageDataCollector`.

packages/studio/src/components/AddPageButton/BasicPageDataCollector.tsx
```diff
@@ -41,6 +42,11 @@ export default function BasicPageDataCollector({
...
+        pageDataValidator.validate(
+          { ...data, url: getPathValue?.value },
+          isEntityPage
```
tmeyer2115, line 48: I'm thinking we should have `isEntityPage` as a variable to the `PageDataValidator()` constructor. It makes sense to me that we'd scope an instance of the validator to only Entity Pages or only Static Pages.

packages/studio/src/utils/PageDataValidator.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * PageDataValidator contains various static utility methods
+ * for validation of user-inputted page data.
+ */
+export default class PageDataValidator {
+  /**
+   * Throws an error if the user-inputted page data is invalid.
+   */
+  validate(
```
tmeyer2115, line 9: For better usability, I don't think we should throw an Error on the first problem we encounter. Suppose a user has both a bad page name and a bad url. They'd first fix one error, try and confirm, then get a new, second error pop up. If we show them all the issues at once, they can fix them all at one.

ann4bel: Updated! I also thought the error messages were a bit crowded so I shortened them a bit: 

Screen Shot 2023-08-07 at 3 25 04 PM

ann4bel: To clarify, the error messages used to include the actual page name in the error message and have "error adding page:" in front, but I removed these. LMK if we should add these back.

packages/studio/src/components/AddPageButton/BasicPageDataCollector.tsx
```diff
@@ -22,6 +23,9 @@ export default function BasicPageDataCollector({
   );
   const { state } = useContext(AddPageContext);
   const isEntityPage = isPagesJSRepo && !state.isStatic;
+  const pageDataValidator = useMemo(() => {
```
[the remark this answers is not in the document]
ann4bel: works! updated to your version

packages/studio/src/components/PageSettingsButton/StaticPageModal.tsx
```diff
@@ -21,7 +22,20 @@ export default function StaticPageModal({
...
+  const [errorMessage, setErrorMessage] = useState<string>("");
+
+  const isPathEditable = useMemo(() => {
+    if (!currGetPathValue) return false;
+    try {
+      pageDataValidator.validate({ url: currGetPathValue.value });
+    } catch (error) {
+      return false;
+    }
+    return true;
```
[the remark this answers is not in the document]
ann4bel: Changed to `isURLEditable`. For the logic extraction, I created a new instance method in `PageDataValidator` that checks whether the url is editable or not.

tmeyer2115 (commented):
Let's sync up today to discuss `PageDataValidator`. I have a couple changes in mind and it's probably easier to discuss via meeting than PR comments! I'll slap something on your calendar.

packages/studio/src/utils/PageDataValidator.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Throws an error if the page name is invalid.
+   */
+  private validatePageName(pageName: string) {
+    const errorMessages: string[] = [];
+    if (!pageName) {
+      errorMessages.push("A page name is required.");
+    }
+    const errorChars = pageName.match(/[\\/?%*:|"<>]/g);
```
jwartofsky-yext, line 51: This will still allow these characters through. Do we want to filter more of them out?

[]{}()!@#$^&_+-=,.;'`~😂

packages/studio/src/utils/PageDataValidator.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * Throws an error if the URL Slug is invalid.
+   */
+  private validateURLSlug(input: string, isEntityPage?: boolean) {
+    const cleanInput = isEntityPage
+      ? input.replace(/\${document\..*?}/g, "")
+      : input;
+    const blackListURLChars = new RegExp(/[ <>""''|\\{}[\]]/g);
```
jwartofsky-yext, line 75: Similar to above, but these characters are allowed through:
!@#$%^&*()_+-=/,.();:`~😂

jwartofsky-yext (approved):

packages/studio/src/utils/PageDataValidator.ts
```diff
@@ -0,0 +1,87 @@
+export interface ValidationResult {
+  valid: boolean;
+  errorMessages: string;
```
tmeyer2115, line 3: I think we should have this be an array of strings. Then, the consumer of this class can choose how it wants to format that. I don't think we should prescribe formatting in this class.

jwartofsky-yext (approved):

tmeyer2115 (approved):

jwartofsky-yext (approved):

PR #303, opened 2023-07-31
Shift Key Allows You To Select Multiple Components

ann4bel:
This pull request adds a shift key macro that allows the user to select multiple adjacent components in the component tree to reorder or delete. The shift key can only be used to select components at the same depth. If the user selects outside the depth, then all descendants in the branch will be selected. 

In addition, all selected components will be highlighted in the preview panel, but the first component that is selected is still considered the active component and will have its props displayed in the prop editor. 

To whoever picks this up: Behind the hood in ComponentTree, if multiple items are dragged and dropped, the way it works is that one item is actually dragged and dropped and the other items follow it. 

There is one minor visual issue (shown in the following Loom) where the red placeholder can show up between two selected components when dragging them, even though dropping them there will not change anything. This functionality is provided by the react-dnd-treeview library, and as such is not easily fixable without submitting a feature request to the library, so we have decided to leave it as-is for now.
https://www.loom.com/share/fbb095ca96bb4b0783b4b6e0b2d3b328

J=SLAP-2838
TEST=auto


packages/studio/src/components/Highlighters.tsx
```diff
@@ -1,20 +1,54 @@
...
-import useActiveComponentName from "../hooks/useActiveComponentName";
+import { CSSProperties, useMemo } from "react";
+import { useComponentNames } from "../hooks/useActiveComponentName";
 import useStudioStore from "../store/useStudioStore";
 import { ComponentStateKind } from "@yext/studio-plugin";
+import DOMRectProperties from "../store/models/DOMRectProperties";
 
 /**
- * Highlights the current activeComponentRect on the page.
+ * Generates Highlighters for all selected components.
```
ann4bel, line 8: Just realized that this would cause overlapping name labels - working on fixing rn!

e2e-tests/tests/add-entity-page.spec.ts
```diff
@@ -20,6 +20,7 @@ studioTest("can add an entity page", async ({ page, studioPage }) => {
   await expect(pageInTree).toHaveCount(1);
   await expect(page).toHaveScreenshot();
   await studioPage.saveButton.click();
+  await page.waitForTimeout(500);
```
ann4bel, line 23: Added this because the test was failing on my personal computer because the local test generation was taking a while, LMK if it should be removed!

oshi97:
I think there are some cases where drag and dropping the active component doesn't work anymore?
https://github.com/yext/studio/assets/23005393/4e8a3da8-9cd2-4ef7-8ec8-37e9fadf5c8f

ann4bel:
> I think there are some cases where drag and dropping the active component doesn't work anymore?
> 
>  Screen.Recording.2023-08-09.at.12.19.49.PM.mov

alextaing:
This might have been what you were talking about in your PR description, but when you do drag components into themselves, they disappear!

https://www.loom.com/share/bf88d1c65e0f47c6a17b0db6abf8c4a6

packages/studio-plugin/src/utils/ComponentTreeHelpers.ts
```diff
@@ -186,4 +186,44 @@ export default class ComponentTreeHelpers {
...
+  ) {
+    const LCAComponent = ComponentTreeHelpers.getLowestCommonAncestorComponent(
+      targetOneUUID,
+      targetTwoUUID,
+      componentTree
+    );
+    const TargetIsLCA =
+      targetOneUUID === LCAComponent?.uuid ||
+      targetTwoUUID === LCAComponent?.uuid;
+    return TargetIsLCA ? LCAComponent?.parentUUID : LCAComponent?.uuid;
```
nmanu1, line 207: instead of doing this logic to get the parent if the LCA is one of the targets, can we just pass the parent of each target to getLowestCommonAncestorComponent to begin with?

packages/studio-plugin/src/utils/ComponentTreeHelpers.ts
```diff
@@ -186,4 +186,44 @@ export default class ComponentTreeHelpers {
...
     return descendants;
   }
+
+  /**
+   * Returns the component UUID that is closest common parent of the two target components.
+   * This is similar to the lowest common ancestor except it cannot be one of the targets.
+   */
+  static getLowestParentUUID(
+    targetOneUUID: string | undefined,
+    targetTwoUUID: string | undefined,
```
nmanu1, line 196: when would either of these UUIDs be undefined?

packages/studio-plugin/src/utils/ComponentTreeHelpers.ts
```diff
@@ -186,4 +186,44 @@ export default class ComponentTreeHelpers {
...
+  }
+
+  private static getLowestCommonAncestorComponent(
+    targetOneUUID: string | undefined,
+    targetTwoUUID: string | undefined,
+    componentTree: ComponentState[]
+  ): ComponentState | undefined {
+    let LCA_UUID;
+    ComponentTreeHelpers.mapComponentTree(componentTree, (c) => {
+      if (!LCA_UUID && targetOneUUID === targetTwoUUID) {
```
nmanu1, line 217: once we find the LCA, we don't need to continue mapping, right? I think it would be cleaner if we short-circuit early if LCA_UUID is defined

packages/studio-plugin/src/utils/ComponentTreeHelpers.ts
ann4bel: when it is the root (which i guess would never happen normally but when I start passing in the parentUUIDs then it will be needed)

packages/studio-plugin/src/utils/ComponentTreeHelpers.ts
ann4bel: is there a way to short circuit out of mapComponentTree?

packages/studio-plugin/src/utils/ComponentTreeHelpers.ts
ann4bel: oh genius! yes changed

packages/studio-plugin/src/utils/ComponentTreeHelpers.ts
ann4bel: actually I just realized that without the undefined, putting `selectedComponentUUIDs.at(0)` as an argument gives an error, so I added it back

packages/studio-plugin/src/utils/ComponentTreeHelpers.ts
nmanu1: oh I just meant returning early inside this callback

ann4bel: oh gotcha! added

ann4bel:
> This might have been what you were talking about in your PR description, but when you do drag components into themselves, they disappear!
> 
> https://www.loom.com/share/bf88d1c65e0f47c6a17b0db6abf8c4a6

Definitely not haha, thanks for the catch! Should be fixed now.

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -160,6 +160,82 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+        throw new Error(
+          `Error adding components: component tree is undefined.`
+        );
+      }
+      if (!activeComponentUUID) {
+        throw new Error(
+          `Error adding components: active component is undefined.`
+        );
+      }
+      const targetComponentUUIDs = [selectedUUID, activeComponentUUID] as [
```
alextaing, line 206: I think this syntax is a little funky?  If you want to keep it as a tuple so you can use `includes` and the spread operator later, I think you can set the type to be `targetComponentUUID: [string, string]`, but IDK if that's preferable (@oshi97/@nmanu1).  You can also just keep them separate, so you don't have do any extra type stuff at all!

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -160,6 +160,82 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+        if (c.parentUUID !== lowestParentUUID) return;
+        const descendants = ComponentTreeHelpers.getDescendants(
+          c,
+          componentTree
+        );
+        const targetIsParent = targetComponentUUIDs.includes(c.uuid);
+        const targetInDescendants = descendants.some((d) =>
+          targetComponentUUIDs.includes(d.uuid)
+        );
+        if (targetIsParent || targetInDescendants) {
```
alextaing, line 226: does this outside check do anything?

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -160,6 +160,82 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+          if (selected || (targetIsParent && targetInDescendants)) {
+            descendants.forEach((d) => addSelectedComponentUUID(d.uuid));
+            addSelectedComponentUUID(c.uuid);
+            selected = false;
+          } else selected = true;
+        }
+        if (selected) {
+          descendants.forEach((d) => addSelectedComponentUUID(d.uuid));
+          addSelectedComponentUUID(c.uuid);
```
alextaing, line 235: The inside of this if statement looks identical to the if statement above (if you remove the outside `if`).  I'm wondering if their logic be consolidated any?

packages/studio/src/components/ComponentNode.tsx
```diff
@@ -30,25 +30,52 @@ interface ComponentNodeProps {
...
     invisible: !hasChild,
   });
 
-  const handleClick = useCallback(() => {
-    setActiveComponentUUID(componentState.uuid);
-  }, [componentState.uuid, setActiveComponentUUID]);
+  const handleClick = useCallback(
+    (event: MouseEvent<HTMLDivElement>) => {
+      if (event.shiftKey && activeComponentUUID) {
+        clearSelectedComponents();
```
alextaing, line 62: Since this is in the `if` and `else` I think you can bring this outside the check

packages/studio/src/components/ComponentTree.tsx
```diff
@@ -167,32 +179,119 @@ function useTree(): NodeModel<ComponentState>[] | undefined {
...
+      dragSourceId: string,
+      destinationParentId: string
+    ) => {
+      if (!componentTree) {
+        throw new Error(
+          "Unable to handle drag and drop event: component tree is undefined."
+        );
+      }
+      if (
+        isEqual(
```
alextaing, line 203: could you bring this out into a variable so it look prettier?  : P

packages/studio/src/store/slices/pages/createPageSlice.ts
ann4bel: the shift select function has to select an inclusive range so I'm not sure if it can be consolidated into one. but the similar code can be combined into a function and called twice if that would be better

packages/studio/src/store/slices/pages/createPageSlice.ts
ann4bel: The outside check `targetIsParent || targetInDescendants` checks whether we've reached the beginning or end of the range we need to select (between the active component and the clicked component). The inside check `(targetIsParent && targetInDescendants)` checks whether the start is the parent and the end is one of the children or vice versa because that would mean that we can just add everything in that branch and then stop.

ann4bel: The way I just typed that was quite confusing lol, LMK if you need clarification.

packages/studio/src/components/ComponentTree.tsx
ann4bel: updated!

packages/studio/src/components/ComponentNode.tsx
ann4bel: updated!

packages/studio/src/store/slices/pages/createPageSlice.ts
ann4bel: changed to `targetComponentUUID: [string, string]`! I like having them together bc I think it makes the code cleaner but up to to @oshi97/@nmanu1

ann4bel: talked to nidhi and changed to `const targetComponentUUIDs = [selectedUUID, activeComponentUUID] as const;`

packages/studio-plugin/src/utils/ComponentTreeHelpers.ts
```diff
@@ -175,15 +175,53 @@ export default class ComponentTreeHelpers {
...
+
+  private static getLowestCommonAncestorComponentUUID(
+    targetOneUUID: string | undefined,
+    targetTwoUUID: string | undefined,
+    componentTree: ComponentState[]
+  ): string | undefined {
+    let LCA_UUID;
+    ComponentTreeHelpers.mapComponentTree(componentTree, (c) => {
+      if (LCA_UUID) return;
+      if (!LCA_UUID && targetOneUUID === targetTwoUUID) {
```
nmanu1, line 215: nit: we don't need to explicitly check for `!LCA_UUID` now that we return early if `LCA_UUID` is truthy

packages/studio/src/components/ComponentNode.tsx
```diff
@@ -109,19 +132,27 @@ export default function ComponentNode(props: ComponentNodeProps): JSX.Element {
...
+        store.pages.selectedComponentUUIDs,
+        store.pages.clearSelectedComponents,
+      ];
+    });
   return useCallback(
     (event: KeyboardEvent<HTMLDivElement>) => {
-      if (isActiveComponent && event.key === "Backspace") {
-        removeComponent(componentStateUUID);
+      if (isSelectedComponent && event.key === "Backspace") {
+        selectedComponentUUIDs.forEach((uuid) => removeComponent(uuid));
```
nmanu1, line 147: I think it would make sense to have the `removeComponent` action remove the component from the selected components, just like it handles resetting the `activeComponentUUID` if the active component is deleted. That way, we shouldn't need to separately call `clearSelectedComponents` below

packages/studio/src/components/RemoveElementButton.tsx
```diff
@@ -14,13 +14,33 @@ interface RemoveElementButtonProps {
...
+      store.pages.clearSelectedComponents,
+      store.pages.activeComponentUUID,
+      store.pages.selectedComponentUUIDs,
+    ];
+  });
 
   const handleClick = useCallback(() => {
-    removeComponent(elementUUID);
-  }, [elementUUID, removeComponent]);
+    if (elementUUID !== activeComponentUUID) removeComponent(elementUUID);
```
nmanu1, line 32: I thought RemoveElementButton only gets added for the active component? When would this case occur?

Also, could the active component ever appear in the middle of the range of selected components? In that case, the button would appear in the middle, which could look weird

packages/studio/src/hooks/useActiveComponentName.tsx
```diff
@@ -16,6 +17,27 @@ export default function useActiveComponentName(): string | undefined {
...
+    return [];
+  }
+  const componentStates = ComponentTreeHelpers.mapComponentTreeParentsFirst(
+    componentTree,
+    (c) => {
+      if (uuids.has(c.uuid)) return c;
+      else return undefined;
+    }
+  );
+  return componentStates.flatMap((state) =>
```
nmanu1, line 36: nit: can the `undefined` values be filtered out of `componentStates` before we do this `flatMap`? I think that would be a little more clear

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -160,6 +160,79 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+    addShiftSelectedComponentUUIDs: (selectedComponent: ComponentState) => {
+      const selectedUUID = selectedComponent.uuid;
+      const {
+        activeComponentUUID,
+        getActivePageState,
+        addSelectedComponentUUID,
+      } = get();
+      const activePageState = getActivePageState();
+      const componentTree = activePageState?.componentTree;
+      if (selectedUUID === activeComponentUUID) return;
```
nmanu1, line 195: In this case, shouldn't the active component be set as the only selected component, instead of having no selected components?

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -160,6 +160,79 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+      if (!componentTree) {
+        throw new Error(
+          `Error adding components: component tree is undefined.`
```
nmanu1, line 198: nit: I find this error message and the one below a little misleading. Can they be changed to start with `Error selecting components`?

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -160,6 +160,79 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+      const targetComponentUUIDs = [selectedUUID, activeComponentUUID] as const;
+      const lowestParentUUID = ComponentTreeHelpers.getLowestParentUUID(
+        ...targetComponentUUIDs,
+        componentTree
+      );
+
+      let selecting = false;
+      ComponentTreeHelpers.mapComponentTreeParentsFirst(componentTree, (c) => {
+        if (c.parentUUID !== lowestParentUUID) return;
+        const descendants = ComponentTreeHelpers.getDescendants(
```
nmanu1, line 215: I think the logic in this handler (lines 215-233) can be simplified. Based on my understanding, there are really only a few cases here:
- _the sub-tree starting with the current component includes **both** the targets_: select the whole sub-tree and exit
- _the sub-tree starting with the current component includes **one** target_: select the whole sub-tree and begin/end selecting
- _the sub-tree starting with the current component includes **no** targets_: only select the whole sub-tree if we're in the middle of selecting

packages/studio/src/store/slices/pages/createPageSlice.ts
briantstephan: Done

packages/studio/src/components/ComponentTree.tsx
```diff
@@ -60,24 +57,52 @@ export default function ComponentTree(): JSX.Element | null {
...
+        if (!node.data) {
+          throw new Error(`Node missing data ${JSON.stringify(node, null, 2)}`);
+        }
+        if (
+          isDragging &&
+          selectedComponentUUIDs.size !== 0 &&
+          !selectedComponentUUIDs.has(node.data.uuid)
+        ) {
+          setActiveComponentUUID(undefined);
+          clearSelectedComponents();
```
nmanu1, line 89: I think we'd always want to reset the selected components when the active component changes. Can we update the store so it handles that behavior internally? I think it would be easy to forget to update the selected components in the future

packages/studio/src/components/ComponentTree.tsx
```diff
@@ -167,32 +204,118 @@ function useTree(): NodeModel<ComponentState>[] | undefined {
...
+          return updateComponentParentUUID(c, destinationParentId);
+        });
+      updatedComponentTree = updatedComponentTree.filter(
+        (c) => c.uuid === dragSourceId || !selectedComponentUUIDs.has(c.uuid)
+      );
+      let newDestinationIndex;
+      updatedComponentTree.forEach((c, index) => {
+        if (c.uuid === dragSourceId) {
+          newDestinationIndex = index;
         }
```
nmanu1, line 259: can we use `Array.findIndex` to calculate `newDestinationIndex`?

packages/studio/src/components/ComponentTree.tsx
```diff
@@ -167,32 +204,118 @@ function useTree(): NodeModel<ComponentState>[] | undefined {
...
+      updatedComponentTree: ComponentState[],
+      dragSourceId: string,
+      destinationParentId: string
+    ) => {
+      if (!componentTree) {
+        throw new Error(
+          "Unable to handle drag and drop event: component tree is undefined."
+        );
+      }
+      const hasNoChanges = isEqual(
```
nmanu1, line 227: for my own understanding, what's an example where `handleSelectedDrop` would be called but `hasNoChanges` is true? I see that in `handleDrop` we already check that the destination is not one of the selected components before calling `handleSelectedDrop`

packages/studio/src/components/ComponentTree.tsx
```diff
@@ -167,32 +204,118 @@ function useTree(): NodeModel<ComponentState>[] | undefined {
...
   );
+
+  const handleDrop = useCallback(
+    (tree: NodeModel<ComponentState>[], options) => {
+      const { dragSourceId, destinationIndex } = options;
+      const updatedComponentTree: ComponentState[] =
+        convertNodeModelsToComponentTree(tree);
+      if (selectedComponentUUIDs.has(dragSourceId)) {
+        const destinationParentId = tree[destinationIndex].parent.toString();
+        if (selectedComponentUUIDs.has(destinationParentId)) return;
```
nmanu1, line 279: why do we check if the destination's parent is one of the selected components rather than the destination itself?

packages/studio-plugin/src/utils/ComponentTreeHelpers.ts
briantstephan: Done

packages/studio/src/components/RemoveElementButton.tsx
briantstephan: Done, the X is now only visible when the active component is the first component in the range of selected components.

packages/studio/src/hooks/useActiveComponentName.tsx
briantstephan: Done

packages/studio/src/store/slices/pages/createPageSlice.ts
briantstephan: Done

packages/studio/src/components/ComponentTree.tsx
briantstephan: Done

packages/studio/src/components/ComponentTree.tsx
briantstephan: Done

packages/studio/src/components/ComponentNode.tsx
briantstephan: Done

packages/studio/src/store/slices/pages/createPageSlice.ts
briantstephan: Done

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -158,6 +163,29 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+      const orderedSelectedComponentUUIDs =
+        ComponentTreeHelpers.mapComponentTreeParentsFirst(
+          componentTree,
+          (c) => {
+            if (selectedComponentUUIDs.has(c.uuid)) {
+              return c.uuid;
+            }
+            return undefined;
+          }
+        ).filter((uuid) => !!uuid);
```
nmanu1, line 184: nit: I think the mapping logic could be a little simpler if we filtered to the selected components first

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -158,6 +163,29 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+      return orderedSelectedComponentUUIDs.length !== 0
+        ? orderedSelectedComponentUUIDs[0]
```
nmanu1, line 186: instead of the ternary, can we directly return `orderedSelectedComponentUUIDs[0]` here since it should already be undefined if it doesn't exist?

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -187,10 +215,15 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
         addSelectedComponentUUID,
+        clearSelectedComponents,
       } = get();
       const activePageState = getActivePageState();
       const componentTree = activePageState?.componentTree;
-      if (selectedUUID === activeComponentUUID) return;
+      if (selectedUUID === activeComponentUUID) {
+        clearSelectedComponents();
+        addSelectedComponentUUID(selectedUUID);
+        return;
```
nmanu1, line 225: can we add/update some unit tests, particularly for store actions that were added/changed?

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -158,6 +163,108 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+
+      let selecting = false;
+      ComponentTreeHelpers.mapComponentTreeParentsFirst(componentTree, (c) => {
+        if (c.parentUUID !== lowestParentUUID) return;
+        const descendants = ComponentTreeHelpers.getDescendants(
+          c,
+          componentTree
+        );
+        const targetIsParent = targetComponentUUIDs.includes(c.uuid);
+        const targetInDescendants = descendants.some((d) =>
```
nmanu1, line 251: I still find this logic a little complicated. We don't actually care where the target is present as long as it's in the sub-tree that starts with the current component or not. Could we update the callback so we get the sub-tree (not just descendants) and then use `reduce` to get the number of targets present in the sub-tree? then the handling logic could be simpler, something like
```ts
if (numTargetsInSubTree === 0 && !selecting) {
  return;
}
subTree.forEach((d) => addSelectedComponentUUID(d.uuid));
if (numTargetsInSubTree === 1) {
  selecting = !selecting;
}
```

packages/studio/src/components/Highlighters.tsx
```diff
@@ -0,0 +1,104 @@
...
+function Highlighter({
+  uuid,
+  rect,
+  iframeEl,
+}: {
+  uuid: string;
+  rect: DOMRectProperties | undefined;
+  iframeEl: HTMLIFrameElement | null;
+}) {
+  const componentTree = useStudioStore((store) =>
```
nmanu1, line 51: it looks like `componentTree` is only used to calculate `isErrorState` below? if so, can we get the `componentTree` directly from the store in that selector callback instead of subscribing to updates to it here?

packages/studio/src/components/Highlighters.tsx
```diff
@@ -0,0 +1,104 @@
...
+    };
+  }, [rect, color]);
+
+  if (!rect) {
+    return null;
+  }
+
+  return (
+    <div style={style}>
+      <div style={tagStyle}>{componentName ? componentName.name : ""}</div>
```
nmanu1, line 101: If we couldn't calculate the component name, it might be better to remove the `div` for the tag entirely

packages/studio/src/components/HighlightingContainer.tsx
```diff
@@ -64,12 +81,12 @@ class HighlightingClass extends Component<HighlightingProps> {
...
     }
-    const rect = rectToJson(childNode.getBoundingClientRect());
-    if (!isEqual(this.props.rect, rect)) {
-      this.props.setRect(rect);
-    }
     if (this.props.activeComponentUUID !== this.props.uuid) {
+      const rect = rectToJson(childNode.getBoundingClientRect());
       this.props.setActiveUUID(this.props.uuid);
+      this.props.clearSelectedComponents();
+      this.props.addSelectedComponentUUID(this.props.uuid);
```
nmanu1, line 88: I think lines 87 and 88 are now internally handled by `setActiveComponentUUID` (line 86)?

packages/studio/src/components/HighlightingContainer.tsx
```diff
@@ -87,6 +104,20 @@ class HighlightingClass extends Component<HighlightingProps> {
   }
 
   componentDidUpdate(): void {
+    const childNode = getDOMNode(this);
+    if (!childNode) {
+      return;
+    }
+    const rect = rectToJson(childNode.getBoundingClientRect());
+    const isRectUpdated =
+      JSON.stringify(this.props.selectedComponentRectsMap[this.props.uuid]) ===
```
nmanu1, line 113: nit: can we use Lodash `isEquals` here instead of the `JSON.stringify`?

packages/studio/src/components/HighlightingContainer.tsx
```diff
@@ -87,6 +104,20 @@ class HighlightingClass extends Component<HighlightingProps> {
   }
 
   componentDidUpdate(): void {
+    const childNode = getDOMNode(this);
+    if (!childNode) {
+      return;
+    }
+    const rect = rectToJson(childNode.getBoundingClientRect());
```
nmanu1, line 111: can some of this logic be consolidated with the `highlightSelf` logic? it looks pretty similar except for whether the active component logic

packages/studio/src/components/HighlightingContainer.tsx
```diff
@@ -87,6 +104,20 @@ class HighlightingClass extends Component<HighlightingProps> {
...
+    const isRectUpdated =
+      JSON.stringify(this.props.selectedComponentRectsMap[this.props.uuid]) ===
+      JSON.stringify(rect);
+    if (
+      !isRectUpdated &&
```
nmanu1, line 116: maybe I'm not following correctly, but shouldn't we be checking that `isRectUpdated` is truthy, not falsy?

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -146,7 +146,12 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
 
   const pageComponentActions = {
     setActiveComponentUUID: (activeComponentUUID: string | undefined) => {
```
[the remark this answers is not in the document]
briantstephan: Maybe something like `clearAndSetActiveComponentUUID`? Not sure what would be best.

packages/studio/src/components/ComponentNode.tsx
```diff
@@ -125,34 +128,24 @@ export default function ComponentNode(props: ComponentNodeProps): JSX.Element {
           />
         )}
       </div>
-      {isActiveComponent && (
-        <RemoveElementButton elementUUID={componentState.uuid} />
-      )}
+      {isActiveComponent &&
+        componentState.uuid === firstSelectedComponentUUID && (
```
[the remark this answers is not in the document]
briantstephan: Done

packages/studio/src/hooks/useActiveComponentName.tsx
```diff
@@ -33,9 +33,11 @@ export function useComponentNames(
       else return undefined;
     }
   );
-  return componentStates.flatMap((state) =>
-    state ? { name: getComponentDisplayName(state), uuid: state.uuid } : []
-  );
+  return componentStates
+    .filter((state) => !!state)
+    .flatMap((state) =>
+      state ? { name: getComponentDisplayName(state), uuid: state.uuid } : []
```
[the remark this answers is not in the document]
briantstephan: Done

packages/studio/src/components/ComponentTree.tsx
```diff
@@ -62,24 +62,27 @@ export default function ComponentTree(): JSX.Element | null {
...
       (node: NodeModel<ComponentState>, renderParams: RenderParams) => {
         const { depth, isOpen, hasChild, isDragging } = renderParams;
         if (!node.data) {
           throw new Error(`Node missing data ${JSON.stringify(node, null, 2)}`);
         }
-        if (isDragging && selectedComponentUUIDs.size !== 0 && !selectedComponentUUIDs.has(node.data.uuid)) {
+        if (
+          isDragging &&
+          selectedComponentUUIDs.size !== 0 &&
+          !selectedComponentUUIDs.has(node.data.uuid)
```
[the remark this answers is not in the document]
briantstephan: Removed

packages/studio/src/store/slices/pages/createPageSlice.ts
briantstephan: Done

packages/studio/src/store/slices/pages/createPageSlice.ts
briantstephan: Done. Rather than counting the number of targets in the subtree, I'm checking if either of the targets is in the subtree at all, and I left the check at the beginning that returns if the current parent UUID is not the lowestParentUUID. I think it's a little easier to follow this way but I'm happy to change it if you disagree.

packages/studio/src/components/Highlighters.tsx
briantstephan: Done

packages/studio/src/components/Highlighters.tsx
briantstephan: Done

packages/studio/src/components/HighlightingContainer.tsx
briantstephan: Removed

packages/studio/src/store/slices/pages/createPageSlice.ts
nmanu1: We still need to differentiate between the case where both targets are in the subtree vs. only one. We wouldn't want to start selecting if both targets are in the current subtree because then we'd never stop selecting

briantstephan: I believe this is handled by the `if (c.parentUUID !== lowestParentUUID) return;` check at the beginning. If the parent is the lowestParent, and both targets are in the subtree, then the one must be a child of the other, in which case I believe we would want to select all children of the higher component. I'll switch it to count the number of targets in the subTree since that lets us get rid of lowestParentUUID entirely, but I _think_ both should work.

packages/studio/src/components/ComponentTree.tsx
briantstephan: I'm not sure why that was being checked. The `hasNoChanges` check in `handleSelectedDrop` takes care of it so I'm just going to remove the destination parent check to avoid confusion.

packages/studio/src/components/ComponentTree.tsx
briantstephan: I can't find any specific ones at the moment, but I remember encountering a few edge cases when dragging components around that were only handled by checking `hasNoChanges`. Only checking that the destination is not a selected component is not sufficient, since the "destination" is really the parent in which the components are dropped, not the exact position. I also like the readability `hasNoChanges` and think it makes more sense than checking specific cases.

packages/studio/src/store/slices/pages/createPageSlice.ts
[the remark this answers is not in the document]
briantstephan: Going with updateActiveComponent 👍

packages/studio/src/store/slices/pages/createPageSlice.ts
nmanu1: nvm, I think that could be problematic since `mapComponentTreeParentsFirst` wouldn't necessarily have the correct `parentUUID` to look for within the selected components unless you compute that first and pass it in

packages/studio/src/components/ComponentTree.tsx
nmanu1: I think this still needs to be updated?

packages/studio/src/store/slices/pages/createPageSlice.ts
briantstephan: Added extra tests for updateSelectedComponent and getFirstSelectedComponentUUID

packages/studio/src/store/slices/pages/createPageSlice.ts
briantstephan: I tried simplifying the logic further but all of the very simple solutions had issues with various edge cases. I updated it to handle all the cases I could find, and the flow is roughly as follows:

- `selecting` starts as `false` and only changes when the selected or active components are reached (I think it might have been swapping back and force multiple times previously)
- If the current parent is not the lowest common parent between the selected and active components, return; we never want to select anything higher up in the tree than lowest common parent
- Generate subtree
- If either the subtree contains one of the targets or `selecting` is `true`, then add all elements in the subtree to `selectedComponentUUIDs`

packages/studio/src/components/HighlightingContainer.tsx
briantstephan: `isRectUpdated` was previously checking for equality between the rectangles rather than a change, so it was working properly but the naming was backwards. I moved the `!` to make the variable name more correct.

packages/studio/src/components/HighlightingContainer.tsx
nmanu1: I think the name `highlightSelf` is a little misleading for what that function currently does. To me, it seems to highlight itself and set itself to the active component. I was thinking, we could remove the active component check and setting and possibly move that into a separate function. Then the click handler would call the modified `highlightSelf` and perform the active component checking/setting. And `componentDidUpdate` would call `highlightSelf` if the component is a selected component and it's rect has changed. We could also move the `isRectUpdated` check into `highlightSelf`

e2e-tests/tests/shift-rearrange-elements.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+import { studioTest } from "./infra/studioTest.js";
+import fs from "fs";
+
+const expectedPage = fs.readFileSync(
+  "./tests/__fixtures__/shift-rearrange-elements-expected-page.tsx",
+  "utf-8"
+);
+
+studioTest(
+  "can rearrange multiple elements in tree using shift",
```
nmanu1, line 11: can this test be updated to have a more visible change? right now, the preview panel doesn't show any change from the rearranging

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -158,6 +163,102 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+      const selectedUUID = selectedComponent.uuid;
+      const {
+        activeComponentUUID,
+        getActivePageState,
+        addSelectedComponentUUID,
+        clearSelectedComponents,
+      } = get();
+      const activePageState = getActivePageState();
+      const componentTree = activePageState?.componentTree;
+      if (selectedUUID === activeComponentUUID) {
```
nmanu1, line 220: nit: can this conditional be moved after the `!activeComponentUUID` check? I think it reads a little easier that way

packages/studio/src/components/HighlightingContainer.tsx
briantstephan: I like that structure, just updated it.

packages/studio/src/components/ComponentTree.tsx
nmanu1: I don't think we can remove the destination check because the tree could still have changes if you drag one of the selected components above another selected component (I'm seeing [this](https://github.com/yext/studio/pull/303#issuecomment-1672018548) behavior). But in that case, we wouldn't want to update anything. Can we add a test for this case?

packages/studio/src/store/slices/pages/createPageSlice.ts
briantstephan: Done

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -158,6 +163,102 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
...
+        clearSelectedComponents();
```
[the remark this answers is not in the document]
briantstephan: Done

packages/studio/src/components/HighlightingContainer.tsx
```diff
@@ -40,9 +52,15 @@ export default function HighlightingContainer(
...
-  rect?: DOMRectProperties;
+  updateActiveComponent: (uuid: string) => void;
+  selectedComponentUUIDs: Set<string>;
+  selectedComponentRectsMap: Record<string, DOMRectProperties>;
+  addSelectedComponentUUID: (componentUUID: string) => void;
+  addSelectedComponentRect: (
+    selectedUUID: string,
+    rect: DOMRectProperties
+  ) => void;
+  clearSelectedComponents: () => void;
```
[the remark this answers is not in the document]
briantstephan: Done

packages/studio/src/store/slices/pages/createPageSlice.ts
```diff
@@ -129,7 +131,8 @@ export const createPageSlice: SliceCreator<PageSlice> = (set, get) => {
         );
       }
 
-      get().setActiveComponentUUID(undefined);
+      get().updateActiveComponent(undefined);
+      get().clearSelectedComponents();
```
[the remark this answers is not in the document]
briantstephan: Done

packages/studio/src/components/HighlightingContainer.tsx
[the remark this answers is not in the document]
nmanu1: I think `addSelectedComponentUUID` can be removed too

oshi97 (commented):
Seeing some odd behavior with selecting components

https://github.com/yext/studio/assets/23005393/b2733fdb-069f-458c-9029-231b3245eb57

packages/studio/src/components/ComponentTree.tsx
```diff
@@ -167,32 +196,114 @@ function useTree(): NodeModel<ComponentState>[] | undefined {
...
+      dragSourceId: string,
+      destinationParentId: string
+    ) => {
+      if (!componentTree) {
+        throw new Error(
+          "Unable to handle drag and drop event: component tree is undefined."
+        );
+      }
+      const hasNoChanges = isEqual(
+        ComponentTreeHelpers.mapComponentTreeParentsFirst(
```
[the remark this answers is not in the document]
briantstephan: That doesn't seem to work when moving a component within the same parent.

PR #310, opened 2023-08-02
Components: Broken Scrolling on Preview Panel

asanehisa:
If you had a lot of components or something long, the preview panel scrolling wasn't working properly. This was due to a fix for toolTips to show properly. I moved toolTips out of the iFrame so it can show properly even without manual fixes. 

This change must be merged before SLAP-2858 as the highlighter box requires the iFrame to work properly for panel preview. 

J=SLAP-2857
TEST=manual, auto

Works with iframe scrolling, main window scrolling, error components w/ tooltips at the top

[loom ](https://www.loom.com/share/3928db725dbb4613bc84c061c5f2655f)


asanehisa:
Screenshot 2023-08-04 at 2 43 53 PM
toolTips!

packages/studio/src/components/ErrorComponentPreview.tsx
```diff
@@ -1,14 +1,13 @@
-import { ErrorComponentState } from "@yext/studio-plugin";
 import { useCallback, useMemo, useRef, useState } from "react";
 import { v4 } from "uuid";
 import ErrorBoundary from "./common/ErrorBoundary";
-import { Tooltip } from "react-tooltip";
 
 export default function ErrorComponentPreview(props: {
-  errorComponentState: ErrorComponentState;
+  errorComponentState: string;
```
alextaing, line 6: Could we preserve the input type of the function to stay as type `ErrorComponentState`?  I think we can still pass in the error state and extract the message from the error state within this function!

packages/studio/src/App.tsx
```diff
@@ -13,7 +21,15 @@ export default function App() {
         <div className="flex flex-row grow">
           <LeftSidebar />
           <div className="grow w-1/3 bg-white border-8 shadow">
-            <PreviewWithUseComponents />
+            <Tooltip
```
alextaing, line 24: Since these tooltips are only used for `PreviewWithComponents`, could we move the tooltip rendering logic into the `PreviewWithComponents` component?

packages/studio/src/App.tsx
```diff
@@ -3,8 +3,16 @@ import ActionsBar from "./components/ActionsBar";
 import Toast from "./components/Toast";
 import PreviewWithUseComponents from "./components/PreviewWithUseComponents";
 import LeftSidebar from "./components/LeftSidebar";
+import { useState } from "react";
+import { Tooltip } from "react-tooltip";
 
 export default function App() {
+  const [tooltipProps, setTooltipProps] = useState({
+    open: false,
+    position: { x: 0, y: 0 },
```
alextaing, line 12: Since we use position to determine the location of the tooltip, if we hover and cause a tooltip to appear then scroll, the component will move but the tooltip will not.  I think this issue is similar to your other highlighting scroll item!

asanehisa: [Fixed](https://www.loom.com/share/3928db725dbb4613bc84c061c5f2655f)!

alextaing:
Screenshot 2023-08-07 at 2 05 40 PM

I noticed one more unexpected behavior-- when we scroll the main window (not the iFrame), the tooltip appears in the location it would have been before scrolling

asanehisa:
I believe I rebased - but unsure if I am suppose to have all the changes in files I didn't actually edit (changes r from rebase)

alextaing:
> I believe I rebased - but unsure if I am suppose to have all the changes in files I didn't actually edit (changes r from rebase)

I think that's ok!  I had that before and it seemed to be fine.  When we merge into main, we normally squash and merge so it all goes away and becomes one commit anyways.

packages/studio/src/App.tsx
alextaing: just a quick follow-up-- I haven't tried it out myself, but is it possible to make the tooltip follow the component while you are hovered over it and scrolling instead of disappearing?  Not sure how complicated that would be, but wanted to check if you knew.

oshi97:
merging with main may resolve some issues?

packages/studio/src/components/PreviewWithUseComponents.tsx
```diff
@@ -4,33 +4,47 @@ import PreviewPanel from "./PreviewPanel";
...
-      className="h-full w-full"
-      title="PreviewPanel"
-      inlineStyles={inlineStyles}
-    >
-      <div style={innerIframeStyles}>
-        <PreviewPanel />
-        <div className="absolute top-[-50px]">
-          <Highlighter />
+    <>
+      <Tooltip
```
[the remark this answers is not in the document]
asanehisa: got it. highlight has a z-index of 10 and modal has a z-index of 1000, so I gave it a z-index of 20 as it should be higher than highlight.

packages/studio/src/App.tsx
asanehisa: It is possible I believe! I asked Tom and I believe he wanted it to just disappear on scroll. That seems to be pretty common behavior for react-tooltips.

packages/studio/src/components/PreviewWithUseComponents.tsx
```diff
@@ -11,6 +11,10 @@ const inlineStyles: CSSProperties = {
   overflow: "scroll",
 };
 
+const tooltipStyle: CSSProperties = {
+  zIndex: "20",
```
nmanu1, line 15: can we add the `z-20` tailwind class to the tooltip `className` instead of adding a `style`? generally, we prefer to use tailwind classes for styling where possible

asanehisa: oh also switched it to z-30 as panels had 20

packages/studio/src/components/PreviewWithUseComponents.tsx
```diff
@@ -4,33 +4,47 @@ import PreviewPanel from "./PreviewPanel";
...
+        className="text-sm z-30"
+        isOpen={tooltipProps.open}
+        position={tooltipProps.position}
+      />
+      <IFramePortal
+        className="h-full w-full"
+        title="PreviewPanel"
+        inlineStyles={inlineStyles}
+      >
+        <div>
```
nmanu1, line 41: nit: now that the inner iframe styling is gone, can this div be removed?

packages/studio/src/components/PreviewWithUseComponents.tsx
```diff
@@ -4,33 +4,47 @@ import PreviewPanel from "./PreviewPanel";
...
-      <div style={innerIframeStyles}>
-        <PreviewPanel />
-        <div className="absolute top-[-50px]">
-          <Highlighter />
+    <>
+      <Tooltip
+        offset={-30}
+        content={tooltipProps.error}
+        anchorSelect={tooltipProps.anchorId}
+        className="text-sm z-30"
```
nmanu1, line 32: out of curiosity, what's the reason for using z-30 rather than z-20?

packages/studio/src/components/PreviewWithUseComponents.tsx
nmanu1: which panels have z-20? I thought only our tooltips would have that index

asanehisa: Ah you are right. I misread it, sorry.

nmanu1: no worries! it's not technically a problem, but I just find it easier and cleaner to only have a couple z-indexes to keep track of

packages/studio/src/App.tsx
alextaing: haha one last thing-- this might be a small nit, but if we have the disappearing behavior on the the scroll in the iFrame, should we have that same behavior when we scroll in the main window?  Right now, when scrolling in the main window, the tooltip follows the component.

asanehisa: oooh kk, thanks!

packages/studio/src/components/ErrorComponentPreview.tsx
```diff
@@ -1,14 +1,14 @@
...
 import { useCallback, useMemo, useRef, useState } from "react";
 import { v4 } from "uuid";
 import ErrorBoundary from "./common/ErrorBoundary";
-import { Tooltip } from "react-tooltip";
+import { ErrorComponentState } from "@yext/studio-plugin";
 
 export default function ErrorComponentPreview(props: {
   errorComponentState: ErrorComponentState;
   element: JSX.Element | null;
+  setTooltipProps;
```
nmanu1, line 9: nit: can we add a type here (and other places we have setTooltipProps as a prop)?

nmanu1 (commented):
I checked out your branch locally and when I switch to a page that has an error component on it, I see this error in the console
Screen Shot 2023-08-08 at 3 54 36 PM

packages/studio/src/components/ErrorComponentPreview.tsx
asanehisa: fixed + done!

packages/studio/src/components/PreviewWithUseComponents.tsx
```diff
@@ -4,33 +4,45 @@ import PreviewPanel from "./PreviewPanel";
...
+import { useState } from "react";
+import { Tooltip } from "react-tooltip";
 
 const inlineStyles: CSSProperties = {
-  transform: "translateY(-50px)",
-};
-
-const innerIframeStyles: CSSProperties = {
-  transform: "translateY(50px)",
+  overflow: "scroll",
```
nmanu1, line 11: without this style, do we cut off the iframe at the screen height so there is no vertical scrolling?

packages/studio/src/components/PreviewWithUseComponents.tsx
```diff
@@ -4,33 +4,45 @@ import PreviewPanel from "./PreviewPanel";
...
-        <PreviewPanel />
-        <div className="absolute top-[-50px]">
+    <>
+      <Tooltip
+        offset={-30}
+        content={tooltipProps.error}
+        anchorSelect={tooltipProps.anchorId}
+        className="text-sm z-20"
+        isOpen={tooltipProps.open}
+        position={tooltipProps.position}
```
nmanu1, line 34: nit: can we spread the tooltipProps here instead of individually listing out each one?

packages/studio/src/components/PreviewWithUseComponents.tsx
```diff
@@ -4,33 +4,45 @@ import PreviewPanel from "./PreviewPanel";
...
-const innerIframeStyles: CSSProperties = {
-  transform: "translateY(50px)",
+  overflow: "scroll",
 };
 
 export default function PreviewWithUseComponents() {
   const componentTree = useStudioStore((store) =>
     store.actions.getComponentTree()
   );
+  const [tooltipProps, setTooltipProps] = useState({
```
nmanu1, line 18: can we pass the type of Tooltip's props as the generic type param for the useState call here? that way, you'd be able to simplify the type of setTooltipProps in the other files to just accept any of the Tooltip props

packages/studio/src/components/ErrorComponentPreview.tsx
```diff
@@ -23,10 +41,40 @@ export default function ErrorComponentPreview(props: {
   }, []);
 
   const rect = containerRef.current?.getBoundingClientRect();
-  const tooltipPosition = rect && {
-    x: (rect.right + rect.left) / 2,
-    y: rect.top - 40,
-  };
+  const tooltipPosition = useMemo(() => {
```
nmanu1, line 44: it looks like tooltipPosition is only used when setting the tooltip props. can we move this calculation into the useEffect?

packages/studio/src/components/ErrorComponentPreview.tsx
```diff
@@ -23,10 +41,40 @@ export default function ErrorComponentPreview(props: {
...
+    );
+  }, [rect]);
+
+  document
+    .querySelector("iframe")
+    ?.contentDocument?.addEventListener("scroll", () => {
+      setTooltipOpen(false);
+    });
+
+  document.addEventListener("scroll", () => {
```
nmanu1, line 60: nit: I slightly prefer the simplified syntax for the callback (`() => setTooltipOpen(false)`) here and above. it makes it more clear to me that this is the only action occurring

alextaing:
Screenshot 2023-08-09 at 9 10 24 AM
image


If you have a component that is scrolled up past the window, the tooltip can be placed too high such that it gets cut off.  Is it possible to set the Y position to a certain min/max so it maxes out at just over the top of the preview window?

packages/studio/src/components/PreviewWithUseComponents.tsx
asanehisa: No. I don't think that style is doing anything, removed.

packages/studio/src/components/PreviewWithUseComponents.tsx
```diff
@@ -3,24 +3,27 @@ import useStudioStore from "../store/useStudioStore";
...
 import IFramePortal from "./IFramePortal";
-import { CSSProperties } from "react";
+import { Tooltip, IPosition } from "react-tooltip";
 import { useState } from "react";
-import { Tooltip } from "react-tooltip";
 
-const inlineStyles: CSSProperties = {
-  overflow: "scroll",
-};
+export interface tooltipProps {
```
nmanu1, line 9: can we directly use the ITooltip type that's exported from react-tooltip instead of creating our own?

asanehisa: Ohhh yes

asanehisa:
Screenshot 2023-08-09 at 12 05 56 PM
maxed height ^

packages/studio/src/components/ErrorComponentPreview.tsx
```diff
@@ -22,11 +31,27 @@ export default function ErrorComponentPreview(props: {
...
-  const rect = containerRef.current?.getBoundingClientRect();
-  const tooltipPosition = rect && {
-    x: (rect.right + rect.left) / 2,
-    y: rect.top - 40,
-  };
+  document
+    .querySelector("iframe")
+    ?.contentDocument?.addEventListener("scroll", () => setTooltipOpen(false));
+
+  document.addEventListener("scroll", () => setTooltipOpen(false));
```
nmanu1, line 38: can we add these listeners in a useEffect hook so that they're only added on initial render (not on each render) and so they can be removed when un-mounting the component?

packages/studio/src/components/ErrorComponentPreview.tsx
```diff
@@ -31,13 +31,14 @@ export default function ErrorComponentPreview(props: {
...
-
-  document.addEventListener("scroll", () => setTooltipOpen(false));
-
   useEffect(() => {
+    document
+      .querySelector("iframe")
+      ?.contentDocument?.addEventListener("scroll", () =>
+        setTooltipOpen(false)
+      );
+    document.addEventListener("scroll", () => setTooltipOpen(false));
```
nmanu1, line 40: With the dependency array of this `useEffect`, new listeners would still be added any time one of those dependencies (`tooltipOpen`, `anchorId`, etc.) changed. I think these listeners should be in their own `useEffect` with an empty dependency array so they're only added on initial render, unless there's a reason they need to be updated after that. To remove a listener on unmount, you'll need to return a callback that removes the listener (see an example in `useWindowListener.js` in the React [docs](https://react.dev/reference/react/useEffect#custom-usewindowlistener-hook))

asanehisa: Oh I see

nmanu1: I think you still need to add the removal logic I mentioned?

packages/studio/src/components/ErrorComponentPreview.tsx
```diff
@@ -22,11 +31,38 @@ export default function ErrorComponentPreview(props: {
...
 
-  const rect = containerRef.current?.getBoundingClientRect();
-  const tooltipPosition = rect && {
-    x: (rect.right + rect.left) / 2,
-    y: rect.top - 40,
-  };
+  useEffect(() => {
+    document
+      .querySelector("iframe")
+      ?.contentDocument?.addEventListener("scroll", () =>
```
nmanu1, line 37: nit: can the callback for the listener be pulled out into a function so we can reuse it for adding and removing both listeners?

packages/studio/src/components/ErrorComponentPreview.tsx
```diff
@@ -22,11 +31,36 @@ export default function ErrorComponentPreview(props: {
     setTooltipOpen(false);
```
nmanu1, line 31: actually, could we just rename this function to `closeTooltip` or something so it can be used for `onMouseLeave` and the listeners below?

asanehisa: Yup can do, was debating that haha

packages/studio/src/components/ErrorComponentPreview.tsx
```diff
@@ -1,32 +1,63 @@
...
   }, []);
-  const onMouseLeave = useCallback(() => {
-    setTooltipOpen(false);
+  const onMouseLeave = useCallback(closeTooltip, []);
+
+  useEffect(() => {
+    document
+      .querySelector("iframe")
+      ?.contentDocument?.addEventListener("scroll", closeTooltip);
+    document.addEventListener("scroll", closeTooltip);
```
nmanu1, line 37: when would this listener's callback have an effect? in order for the scroll to occur on the top-level document rather than the iframe, the mouse would have to be outside the iframe, in which case the tooltip would already be closed because of the onMouseLeave, right?

asanehisa: Oh so u can oddly scroll in the outer window even while ur in the preview panel.

packages/studio/src/components/ErrorComponentPreview.tsx
```diff
@@ -1,32 +1,63 @@
...
+    const tooltipPosition = rect && {
+      x: (rect.right + rect.left) / 2 + window.innerWidth / 4 - window.scrollX,
+      y: y < 25 ? 25 : y,
+    };
+
+    setTooltipProps({
+      isOpen: tooltipOpen,
+      content: errorComponentState.message,
+      anchorSelect: anchorId,
+      position: tooltipPosition || { x: 0, y: 0 },
```
nmanu1, line 58: nit: can we use nullish coalescing operator (`??`) here instead of logical OR? that makes it a little more clear to me when the default value should be used

packages/studio/src/components/ErrorComponentPreview.tsx
nmanu1: ah like if you're already at the bottom of the preview panel and continue scrolling down, or if the preview panel is not scrollable but the outer window is. got it!

nmanu1 (approved):
🎉

PR #312, opened 2023-08-03
Add link to live preview in pages dev server

benlife5:
Spawns a pages development server on studio startup.
Adds a button to the studio page that links to the 
pages development index page.
The button always links to :5174 because that
will typically be the pages dev server port.

J=SLAP-2863
TEST=manual,auto

Manually tested by verifying the button works as expected
and that if the pages dev server or studio subprocesses
crash, it takes down all studio-related processes.
Tested on both Mac and Windows.


packages/studio-plugin/src/startPagesDevelopmentServer.ts
```diff
@@ -0,0 +1,53 @@
...
+        setTimeout(resolve, 10000)
+      });
+    }
+  );
+
+  // kill pages server when studio exits
+  process.on("exit", () => {
+    if (port) {
+      console.log("Studio is shutting down")
+      execSync(`npx kill-port ${port}`)
```
oshi97, line 48: you don't need to exec this, you can directly call import the package and call it

packages/studio-plugin/src/startPagesDevelopmentServer.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  // kill studio if the pages server exits
+  pagesServer.addListener("exit", () => {
+    process.exit(-1);
+  });
+
+  // kill pages server child process when studio exits
+  // by getting the dev server port on startup and then
+  // killing the proceess running at that port
+  const port : number | void = await new Promise(
```
oshi97, line 27: nit but I think I prefer undefined over void here

packages/studio-plugin/src/startPagesDevelopmentServer.ts
benlife5: The issue here (and why I'm still doing `await startPagesDevelopmentServer();` is that `process.on("exit"`... doesn't reliably work with async code. The kill-port package is an async function and reading the port from the server's stdout also has to be async

packages/studio-plugin/src/createStudioPlugin.ts
```diff
@@ -65,6 +66,10 @@ export default async function createStudioPlugin(
     new FileSystemWriter(orchestrator, tsMorphProject)
   );
 
+  if (studioConfig.isPagesJSRepo) {
+    await startPagesDevelopmentServer();
```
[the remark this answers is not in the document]
benlife5: It's about 10 seconds from command -> studio page loads. Not ideal but not terrible, see below for why I'm awaiting this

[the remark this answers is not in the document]
oshi97: I poked around a bit and for me it takes about 4 seconds for `pages dev` to finish (from ~12 seconds to ~16 seconds total wait for studio). I think we can run `pages dev` concurrently by calling `startPagesDevelopmentServer` near the top of `createStudioPlugin` before all of studio's parsing starts, but only awaiting the promise returned by it at the bottom.

packages/studio/src/components/OpenLivePreviewButton.tsx
```diff
@@ -0,0 +1,23 @@
+import { useRef } from "react";
+
+/**
+ * OpenLivePreviewButton is a button that when clicked, opens the
+ * pages development server index page in a new tab.
+ * Port 5173 hardcoded for now as it will be the pages dev port in most cases
+ */
+export default function OpenLivePreviewButton(): JSX.Element | null {
+  const containerRef = useRef<HTMLDivElement>(null);
```
oshi97, line 9: think this ref is unused?

packages/studio-plugin/src/createStudioPlugin.ts
```diff
@@ -33,6 +35,24 @@ export default async function createStudioPlugin(
   const pathToUserProjectRoot = getProjectRoot(cliArgs);
 
   const studioConfig = await getStudioConfig(pathToUserProjectRoot, cliArgs);
+
+  let pagesDevelopmentPort: Promise<number | undefined>;
```
oshi97, line 39: I think you can avoid the `let` by doing something like this instead
```ts
const pagesDevPromise: Promise<void> =
  studioConfig.isPagesJSRepo ? startPagesDevelopmentServer() : Promise.resolve();
```
you could alternatively make a wrapper function or even pass isPagesJSRepo into startPagesDevelopmentServer (the latter is a bit weird from a high level)

packages/studio-plugin/src/createStudioPlugin.ts
```diff
@@ -33,6 +35,24 @@ export default async function createStudioPlugin(
...
+    pagesDevelopmentPort = startPagesDevelopmentServer();
+    pagesDevelopmentPort
+      .then((port) => {
+        if (port) {
+          process.on("exit", () => {
+            console.log("Studio is shutting down");
+            execSync(`npx kill-port ${port}`);
+          });
+        }
+      })
```
oshi97, line 50: I think I would prefer for this `.then(() => ...` logic to be inside of startPagesDevelopmentServer?

packages/studio-plugin/src/createStudioPlugin.ts
```diff
@@ -33,6 +35,24 @@ export default async function createStudioPlugin(
...
+      .catch((e) => {
```
oshi97, line 51: The .catch prevents unhandled rejections which is nice! But I think I'd still like to await the promise itself at the end just to avoid potential weird edge cases with studio starting before `pages dev` is done. If we await the promise I think we don't need this .catch anymore.

e2e-tests/tests/infra/spawnStudio.ts
```diff
@@ -23,7 +23,7 @@ export default async function spawnStudio(
   const port = await getPort(testInfo);
   const child = spawn(
     "npx",
-    ["studio", "--port", port.toString(), "--root", rootDir],
+    ["studio", "--port", port.toString(), "--root", rootDir, "--", "--noGenerateTestData"],
```
oshi97: right now studio should never try to generate test data on startup, mostly because we want it to run in CBD (i.e. the code editor's live preview) but also because we're still trying to limit the amount of PagesJS specific logic

packages/studio-plugin/src/startPagesDevelopmentServer.ts
```diff
@@ -0,0 +1,62 @@
...
+      console.log(matches);
+      if (matches && matches.length >= 2) {
+        const port = Number.parseInt(matches[1]);
+        if (isNaN(port)) {
+          throw new Error("PagesJS dev server port was parsed as NaN.");
+        }
+        resolve(port);
+      }
+    });
+    setTimeout(resolve, 10000);
```
oshi97, line 53: I think this will resolve to undefined, and not a number?

benlife5: Yes, this is if it can't read the port from the terminal for some reason. Do we want it to default to a number or undefined?

oshi97: I would probably throw an error here instead

e2e-tests/tests/infra/spawnStudio.ts
```diff
@@ -23,7 +23,15 @@ export default async function spawnStudio(
...
     "npx",
-    ["studio", "--port", port.toString(), "--root", rootDir],
+    [
+      "studio",
+      "--port",
+      port.toString(),
+      "--root",
+      rootDir,
+      "--",
+      "--noGenerateTestData",
```
oshi97, line 33: could we remove these since these args don't exist anymore

benlife5: Yep

oshi97 (approved):
lgtm!

PR #317, opened 2023-08-04
Generate test data when StreamScope is edited

oshi97:
When stream scope is edited for an entity page, regenerate the test data.

Simplifies the updateActivePageEntities action by moving the dynamic importing logic into StudioActions, which removes
the need to manually pass in the localDataFolder.

J=SLAP-2848
TEST=manual

when stream scope is updated, test data is regenerated and the page uses the new data
when an error occurs, get a warning toast and the settings are still updated


packages/studio/src/components/PageSettingsButton/EntityPageModal.tsx
```diff
@@ -64,9 +77,28 @@ export default function EntityPageModal({
         updateGetPathValue(pageName, getPathValue);
       }
       updateStreamScope(pageName, StreamScopeParser.parseStreamScope(form));
+      try {
+        const mapping = await generateTestData();
```
tmeyer2115, line 81: Could we guard this behavior by first checking if the Stream Scope was edited? If the Scope wasn't modified (maybe it was just the URL), we shouldn't need to regenerate `localData` or `refreshActivePageEntities`, right?

packages/studio/src/store/models/slices/PageSlice.ts
```diff
@@ -68,7 +68,9 @@ interface PageSliceActions {
   setModuleUUIDBeingEdited: (moduleStateUUID: string | undefined) => void;
 
   setActiveEntityFile: (activeEntityFile?: string) => void;
-  updateActivePageEntities: (parentFolder: string) => Promise<void>;
+  updateActivePageEntities: (
```
tmeyer2115, line 71: Could we call this `setActivePageEntities`? By moving the import logic out, we've effectively just made this method a setter. I think the `update` verbiage implies some extra work before the `set` is called.

packages/studio/src/components/PageSettingsButton/EntityPageModal.tsx
```diff
@@ -23,13 +24,25 @@ export default function EntityPageModal({
...
+    refreshActivePageEntities,
+  ] = useStudioStore((store) => [
+    store.pages.pages[pageName].pagesJS?.getPathValue,
+    store.pages.updateGetPathValue,
+    store.pages.pages[pageName].pagesJS?.streamScope,
+    store.pages.updateStreamScope,
+    store.actions.generateTestData,
+    store.pages.updateEntityFiles,
+    store.pages.setActiveEntityFile,
+    store.actions.refreshActivePageEntities,
```
tmeyer2115, line 44: Outside the scope of this PR, but I think we should re-examine our Slice design. Specifically, I'm thinking we should have a dedicate Entity Data Slice in the Store. That would be responsible for maintaining an up-to-date mapping of page name to a set of entities. In the `pages` Slice, we'd only have a method for setting the Active Entity used for Previews (`setActiveEntityFile`). I think this could lead to better separation of concerns.

oshi97: the way zustand does slices is a little restrictive, but I think definitely the store can use some reorganizing

packages/studio/src/components/PageSettingsButton/EntityPageModal.tsx
```diff
@@ -64,9 +77,33 @@ export default function EntityPageModal({
...
+          setActiveEntityFile(mapping[pageName]?.[0]);
+          await refreshActivePageEntities();
+        } catch {
+          toast.warn(
+            "Error generating test data, but entity page settings were still updated."
+          );
+        }
+      };
+      if (form.entityTypes || form.entityIds || form.savedFilterIds) {
+        await regenerateTestData();
```
nmanu1, line 93: wouldn't we want to generate test data any time even if the stream scope isn't filtered to anything?

oshi97: I updated the guarding logic it wasn't what I intended

tmeyer2115 (approved):

PR #318, opened 2023-08-07
Shorten git workflows

alextaing:
This PR reduces some unnecessary GH actions checks. First linting is performed. If linting fails, tests will not be run. If at any time new commit is pushed to a PR while the old workflow is running, the old workflow will be cancelled and replaced by the new one.

Manual testing was performed by merging in branches into this branch with GitHub checks.  Workflows from the same PR get cancelled when new commits are pushed.  Pushes to different PRs do not affect each other.

J=SLAP-2873
TEST=manual


.github/workflows/root-workflow.yml
```diff
@@ -0,0 +1,24 @@
+name: Root Workflow
+
+on: pull_request
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  automated-fixes:
```
oshi97, line 10: Is there some way we can hide the `Root Workflow` prefix from here? Or just shorten it a little
image
Since they all have Root Workflow on them now it's a little redundant
I noticed the SDK PR checks don't have a prefix despite having a similar flow but I can't quite figure out what's different about theirs
https://github.com/yext/answers-search-ui/pull/1832

.github/workflows/root-workflow.yml
```diff
@@ -0,0 +1,24 @@
...
+  automated-fixes: 
+    uses: ./.github/workflows/autofixes.yml
+    secrets: inherit
+
+  update-tailwind-bundle: 
+    uses: ./.github/workflows/update-tailwind-bundle.yml
+    secrets: inherit
+
+  run-tests:
+    needs: automated-fixes
```
oshi97, line 19: We want to prevent our tests from running not only when automated-fixes fails but ALSO when it succeeds but puts up a commit
I think the easiest way would be to fail the autofixes workflow if a commit was made? Not sure exactly how you'd do that, but I think you can use the output from the add-and-commit action
https://github.com/EndBug/add-and-commit#outputs

alextaing: I think that is taken care of by the `concurrency` above!  Whenever another commit is pushed to a PR, it will cancel everything (automated fixes, unit tests, playwright tests, tw bundling) and only run the new checks.

.github/workflows/root-workflow.yml
alextaing: I've done some digging in their workflows and online and for some reason I can't find how they do it. For now I'll shorten it to `run`

alextaing: changed to `main`

oshi97 (dismissed):

.github/workflows/main-workflow.yml
```diff
@@ -0,0 +1,25 @@
+name: main  
+run-name: Main Workflow
```
nmanu1, line 2: What's the reason for using `run-name` here? I thought having different runs of this action have different names would be useful for debugging?

alextaing: ahh, I see yeah thats a good point.  I didn't realize that it changed all the actions in the actions tab to have the same name. I'll get rid of it

.github/workflows/playwright.yml
```diff
@@ -1,9 +1,7 @@
 name: Playwright Tests
 
 on:
-  push:
-    branches: main
-  pull_request:
+  workflow_call:
```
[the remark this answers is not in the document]
alextaing: Ill give it a shot!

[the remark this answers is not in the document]
oshi97: should work its just a yml thing!

.github/workflows/autofixes.yml
```diff
@@ -17,32 +20,19 @@ jobs:
...
         uses: EndBug/add-and-commit@v9
         with:
           message: "Automated linting update and features.json sync"
           add: "*"
           push: true
           pull: --rebase --autostash
           default_author: github_actions
+      - name: Fail on linting push
+        run: |
+          if [[ ${{ steps.automated-linting.outputs.pushed }} == true ]]; then exit 1; else exit 0; fi
```
nmanu1, line 33: is this check needed? I thought all the workflows would be cancelled automatically if a new commit was made because cancel-in-progress is true in main-workflow

.github/workflows/autofixes.yml
```diff
@@ -17,32 +20,19 @@ jobs:
...
+
+          
+
+
```
nmanu1, line 37: nit: looks like there are some extra newlines here

.github/workflows/autofixes.yml
alextaing: It's not completely necessary-- I talked with Oliver and he said that it might be nice to have so that the tests don't even start running.  But yeah, when a new push is made it'll cancel the current tests even if we don't have this.  Are you OK with keeping the check here, or would you rather have it removed?

.github/workflows/autofixes.yml
alextaing: oops how'd you guys get there

nmanu1: you should be able to run prettier on the `.github` directory and it'll take care of it. although you'll need to temporarily remove the directory from the `.prettierignore`

oshi97: I think we only had it in the prettierignore before because we didn't have the repo_scoped_bot_token set up, also down to remove it!

PR #320, opened 2023-08-07
v0.17.0.

tmeyer2115:
## Features
- For Entity Templates, you can now pick which Entity is used in the preview. (#311)
- Props can now have Developer-specified display names. When specified, these appear in the Prop Editor in place of the variable name. (#307)
- The Delete key now removes the currently selected Component. (#293)
- When an Entity Template's Stream Scope is edited, we re-generate the Test Data. (#317)

## Changes
- Repeaters are now deprecated. You cannot enable or disable Repeaters in the Studio UI. (#290)
- A Prop's Name now appears above its value selector. (#302)

## Bug Fixes
- Re-generation of `localData` now works properly if you have multiple, unsaved Entity Templates. (#304)
- Re-generation of `localData` no longer can cause a full page refresh. (#304)
- Studio now generates unique Stream Ids for the Entity Templates it creates. (#314)
- Assorted UI fixes. (#288, #298)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23141,7 +23141,7 @@
     },
     "packages/studio": {
       "name": "@yext/studio",
-      "version": "0.16.0",
+      "version": "0.17.0",
       "dependencies": {
         "@dhmk/zustand-lens": "^2.0.5",
         "@minoru/react-dnd-treeview": "^3.4.1",
@@ -23190,7 +23190,7 @@
     },
     "packages/studio-plugin": {
       "name": "@yext/studio-plugin",
-      "version": "0.16.0",
+      "version": "0.17.0",
       "dependencies": {
         "@babel/parser": "^7.21.8",
         "@babel/preset-env": "^7.22.4",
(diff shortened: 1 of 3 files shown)


oshi97:
#304 also fixes a bug where generating the test data can cause a full page refresh, I was thinking we'd include it in the notes since it was something Andrew asked about

oshi97:
I also just merged in https://github.com/yext/studio/pull/317 if we want to include it?

oshi97:
> > #304 also fixes a bug where generating the test data can cause a full page refresh, I was thinking we'd include it in the notes since it was something Andrew asked about
> 
> #304 is already called out in the release notes! It's the first bullet point under bug fixes.

I was thinking we could explicitly say it fixed this bug since it's conceptually separate from the bullet point we already have

tmeyer2115:
> > > #304 also fixes a bug where generating the test data can cause a full page refresh, I was thinking we'd include it in the notes since it was something Andrew asked about
> > 
> > 
> > #304 is already called out in the release notes! It's the first bullet point under bug fixes.
> 
> I was thinking we could explicitly say it fixed this bug since it's conceptually separate from the bullet point we already have

Gotcha! I added a new bullet under fixes.

oshi97 (approved):

PR #323, opened 2023-08-08
Custom Tailwind Support

oshi97:
This PR adds support for custom tailwind themes.
Currently, we only support customizations for bg-color, text color, and font size.
To do this, we import the user's tailwind.config.ts file into studio-plugin, and generate a new tailwind config based on that one.

We now include tailwind on the page via postcss, instead of pre-generating the entire bundle.
The page has access to classes used by our studio source code, and custom background colors, text colors, and font sizes derived from the user's custom tailwind theme.

J=SLAP-2875
TEST=manual,auto

added acceptance test for custom tailwind bg-color and font size


e2e-tests/tests/__fixtures__/hmr/updated-button-body.tsx
```diff
@@ -9,9 +9,7 @@ export const initialProps: ButtonProps = {
 export default function Button(props: ButtonProps) {
   return (
     <button className={props.className}>
-      <div className="bg-purple-400 text-white">
```
oshi97, line 12: the bg-purple class no longer works since we're not including the full tailwind bundle on the page

packages/studio/tailwind.config.ts
```diff
@@ -1,9 +1,60 @@
...
+const userTailwindTheme: StudioTailwindTheme =
+  (() => {
+    try {
+      const tailwindConfigPath = path.resolve(rootDir, "tailwind.config.ts");
+      if (fs.existsSync(tailwindConfigPath)) {
+        // We have to use require() instead of a dynamic import because
+        // tailwind does not support async config.
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        const tailwindConfig: StudioTailwindConfig = require(tailwindConfigPath);
+        return tailwindConfig?.theme?.extend;
```
nmanu1, line 31: For now, do we only support extensions to the Tailwind config and not customizations of existing classes?

packages/studio/tailwind.config.ts
```diff
@@ -1,9 +1,60 @@
...
+    ...customColors.flatMap((color) => [`bg-${color}`, `text-${color}`]),
+  ];
+};
 
 export default {
-  content: [path.join(__dirname, "src/**/*.tsx")],
+  content: [
+    path.resolve(__dirname, "src/**/*.{ts,tsx}"),
+    path.resolve(__dirname, "index.html"),
+    path.resolve(rootDir, "src/**/*.{ts,tsx}"),
```
nmanu1, line 54: For now, are we deliberately making the assumption that this is the only allowed content path for the user's Tailwind? Is there a reason for not including whatever content path/glob they have specified?

packages/studio/tailwind.config.ts
oshi97: yup

packages/studio/tailwind.config.ts
oshi97: We can't just directly spread the user's content array into ours because the relative paths won't work, and there's a few different ways to specify content besides just a string filepath
```ts
type ContentConfig =
  | (FilePath | RawFile)[]
  | {
      files: (FilePath | RawFile)[]
      relative?: boolean
      extract?: ExtractorFn | { [extension: string]: ExtractorFn }
      transform?: TransformerFn | { [extension: string]: TransformerFn }
    }
type FilePath = string
type RawFile = { raw: string; extension?: string }
```

But we can restrict what the user is allowed to specify as content via the StudioTailwindConfig type

packages/studio/tailwind.config.ts
```diff
@@ -47,11 +48,20 @@ const generateSafelist = (theme: StudioTailwindTheme): string[] => {
...
+    return path.join(rootDir, filepath);
+  }
+);
+
 export default {
   content: [
     path.resolve(__dirname, "src/**/*.{ts,tsx}"),
     path.resolve(__dirname, "index.html"),
-    path.resolve(rootDir, "src/**/*.{ts,tsx}"),
+    ...transformedUserContent,
```
nmanu1, line 64: can the custom Tailwind Playwright test be updated so it has an example of user-specified content?

oshi97: if the user content doesn't get passed in then none of the styles except safelisted styles would show up

packages/studio/tailwind.config.ts
oshi97: We still need to have a default here for when user's don't have a tailwind config. In the future maybe we can assume they always have one

packages/studio-plugin/src/types/StudioTailwindConfig.ts
```diff
@@ -0,0 +1,19 @@
+import type { Config } from "tailwindcss";
+
+/**
+ * The subset of a tailwind Config supported by Studio.
+ */
+export type StudioTailwindConfig = Omit<Config, "theme" | "content"> & {
+  content: string[];
+  theme: {
+    extend: StudioTailwindTheme;
```
nmanu1, line 9: Should `theme` and `extend` be optional properties?

packages/studio/tailwind.config.ts
nmanu1: this is no longer the case now that we have a default, right?

oshi97: the tailwind test doesn't use the default

packages/studio/tailwind.config.ts
```diff
@@ -1,9 +1,74 @@
...
+    const tailwindConfigPath = path.resolve(rootDir, "tailwind.config.ts");
+    if (fs.existsSync(tailwindConfigPath)) {
+      // We have to use require() instead of a dynamic import because
+      // tailwind does not support async config.
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const tailwindConfig: StudioTailwindConfig = require(tailwindConfigPath);
+      return tailwindConfig;
+    }
+  } catch (e) {
+    console.error(e);
```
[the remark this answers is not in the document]
oshi97: No I can remove the try

packages/studio-plugin/src/types/StudioTailwindConfig.ts
oshi97: will update!

packages/studio/tailwind.config.ts
oshi97: I updated the tailwind test to have specify and use tailwind classes from files outside the default content
the box shadow classes are pretty weird and I don't expect them to be used by the studio/src files

nmanu1 (approved):
🚀

PR #324, opened 2023-08-08
Plugin: Prop Editors Don't Support Negative Numbers

asanehisa:
Added an if else case for negative numbers. A SyntaxKind.NumericLiteral only accepts whole numbers, not negatives due to the operand. 

J=SLAP-2869
TEST=manual, auto 

Attempted using -1's in Cta, Banner, Buttons, and more components via right-side prop editor.

### DIFF
--- a/packages/studio-plugin/tests/parsers/StaticParsingHelpers.test.ts
+++ b/packages/studio-plugin/tests/parsers/StaticParsingHelpers.test.ts
@@ -56,6 +56,7 @@ describe("parseObjectLiteral", () => {
 describe("parseJsxAttributes", () => {
   const propShape: PropShape = {
     title: { type: PropValueType.string, tooltip: "jsdoc", required: false },
+    number: { type: PropValueType.number, required: false },
     nested: {
       type: PropValueType.Object,
       required: false,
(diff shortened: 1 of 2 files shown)


asanehisa:
Screenshot 2023-08-08 at 11 49 24 AM
Error before fixes

packages/studio-plugin/tests/parsers/StaticParsingHelpers.test.ts
```diff
@@ -304,6 +304,22 @@ describe("parseExportAssignment", () => {
   });
 });
 
+it("parsing an negative numeric literal", () => {
+  const { sourceFile } = createTestSourceFile(`const a = { aKey: -1 }`);
```
tmeyer2115, line 308: Small nit, but could this test instead parse a Component invocation with a negative prop value?

tmeyer2115 (dismissed):

packages/studio-plugin/tests/parsers/StaticParsingHelpers.test.ts
```diff
@@ -201,6 +202,23 @@ describe("parseJsxAttributes", () => {
     };
     expect(receivedPropValues).toEqual(expectedPropValues);
   });
+
+  it("parsing an negative numeric literal", () => {
```
nmanu1, line 206: nit: can this test's title match the style of our other ones (verb phrase so that if you start with "it", they usually form a grammatically correct statement)? also super nit: `an` -> `a`

nmanu1 (approved):

PR #325, opened 2023-08-09
Automate third party notices

alextaing:
This PR automates updating third party notices. Since we have a monorepo and use workspaces with multiple `package.json` files that are not all surface level, we cannot directly use the SLAP reusable workflow. 

PR https://github.com/yext/studio/pull/318 removed `.github` from `.prettierignore` to allow the autofixing of workflows.  However, Slapshot Bot is not configured with workflow permissions, and changing these permissions could affect other teams also using this bot.  Therefore, `.github` was re-added to `.prettierignore`.

J=SLAP-2871
TEST=manual


alextaing:
Do we want to keep the exact same formatting for our third party notices as before, or can we[ upgrade to its v2](https://github.com/TobyAndToby/generate-license-file/releases/tag/v2.0.0)?

.prettierignore
```diff
@@ -4,3 +4,4 @@ packages/studio-plugin/tests/__fixtures__/PageFile/noReturnParenthesesPage.tsx
 packages/studio-plugin/tests/__fixtures__/PageFile/repeaterPage.tsx
 packages/studio-plugin/tests/__fixtures__/StudioConfigs/malformed/studio.config.js
 packages/studio/src/tailwind-full.css
+.github
```
alextaing, line 7: image

I think that our PAT for the bot does not allow for editing the workflow.  Because of this, I re-added `.github` to prettier.  Should we update the PAT permissions instead?

nmanu1: I thought your previous PR had a commit where the GH linting action was able to make a commit that updated the styling for a workflow?

alextaing: I'm slightly confused, because it seems to be able to update some of the work flows (see [this automated commit](https://github.com/yext/studio/pull/325/commits/dcffc7fee3c4ded030ec7d919d5319f8d87230cd)), but not others (see [this failed workflow](https://github.com/yext/studio/actions/runs/5811451007/job/15754715175)). and I'm not completely sure why.

alextaing: It looks like it only causes errors when trying to edit the newly added workflows... looking into it

.github/workflows/update-third-party-notices.yml
```diff
@@ -0,0 +1,33 @@
...
+      - run: npm install -g generate-license-file@1
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+          token: ${{ secrets.BOT_REPO_SCOPED_TOKEN }}
+      # Skip post-install scripts here, as a malicious
+      # script could steal the NPM_TOKEN.
+      - run: npm ci --ignore-scripts
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
nmanu1, line 21: since we're not installing any private npm packages, we don't need this npm token. if we remove it, we can just do an npm install without the `--ignore-scripts` flag and the rebuild step because we'd no longer have to worry about exposing the token

.github/workflows/update-third-party-notices.yml
```diff
@@ -0,0 +1,33 @@
+name: Check and Update Third Party Notices
+on: workflow_call
+
+jobs:
+  license-check:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
```
nmanu1, line 10: nit: can we use a more recent node version?

.prettierignore
alextaing: I looked into this with Tom, and it requires editing Slapshot Bot PAT permissions, which might be a little excessive for the ability to lint workflow files.  I think we will keep `.github` in the prettierignore for now.

.github/workflows/update-third-party-notices.yml
alextaing: is node 18 good?

alextaing:
Use [Yen's workaround](https://github.com/yext/chat-headless/blob/main/generate-notices.sh) to get both direct and transitive dependencies for third party notices in a monorepo!

.github/workflows/main-workflow.yml
```diff
@@ -14,6 +14,10 @@ jobs:
     uses: ./.github/workflows/update-tailwind-bundle.yml
     secrets: inherit
 
+  update-thrid-party-notices:
```
oshi97, line 17: I would include this in automated fixes

oshi97: oh wow just kidding it takes 14 minutes to run?? do we know why it's so slow?

.github/workflows/update-third-party-notices.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  license-check:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
+          registry-url: "https://registry.npmjs.org"
+      - run: npm install -g generate-license-file@1
```
oshi97, line 12: I would add this to our package.json. It's not some huge package right?

.github/workflows/update-third-party-notices.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+          token: ${{ secrets.BOT_REPO_SCOPED_TOKEN }}
+      - name: Generate third party notices for studio
+        run: cd packages/studio; ../../generate-third-party-notices.sh
```
nmanu1, line 18: can we use `working-directory` instead of manually cd-ing into the package?

.github/workflows/update-third-party-notices.yml
oshi97: Then it'll also benefit from our caching action. Technically we can still set up separate caching just for this but it seems unnecessary

.github/workflows/update-third-party-notices.yml
oshi97, line 18: working-directory parameter is clearer

.github/workflows/update-third-party-notices.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Generate third party notices for studio-plugin
+        run: cd packages/studio-plugin; ../../generate-third-party-notices.sh
```
oshi97, line 20: you can add a top level script for this that uses turbo for parallelization

oshi97:
> Do we want to keep the exact same formatting for our third party notices as before, or can we[ upgrade to its v2](https://github.com/TobyAndToby/generate-license-file/releases/tag/v2.0.0)?

ask Tom but I would upgrade

.github/workflows/update-third-party-notices.yml
alextaing: nup I can do that!

.github/workflows/update-third-party-notices.yml
alextaing: yep!

alextaing: yep!

generate-third-party-notices.sh
```diff
@@ -0,0 +1,9 @@
+PACKAGE_DIR=$(pwd)
```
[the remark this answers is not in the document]
alextaing: That file is the one that Yen provided that she used to deal with generating notices for monorepos.  In the past, we've had one `package.json` at the root level of the project with its `package.json` and we could easily just run `generate-license-file` to generate the `THIRD PARTY NOTICES` file.  Since there are multiple `package.json` files in subdirectories, we need a slightly different approach.  If we just run `generate-license-file` in the subdirectories, since they don't have access to the `package-lock.json` file, will generate notices for the direct dependencies, but not the transitive dependencies. This script takes the files in the subdirectories and brings them to a level where they can access the `package-lock.json` file so that notices for transitive dependencies can be generated.  I am slightly unclear about how the very last part of that works, but I will ask Yen and get back to you on that.

.github/workflows/main-workflow.yml
alextaing: It normally takes around 2-3 mintues, I'm not sure why it took so long that time

generate-third-party-notices.sh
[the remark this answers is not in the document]
alextaing: I talked with Yen and might try one more approach where I edit Workspaces in the top level `package.json` so we can avoid the copying and pasting stuff.

.github/workflows/update-third-party-notices.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  license-check:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
+          registry-url: "https://registry.npmjs.org"
+      - run: npm install -g json@v11
+      - run: npm install -g generate-license-file@2
```
alextaing, line 13: I think I like the generate license install here!  Otherwise, I would probably need to have it as a dev dependency in both studio and studio-plugin.  I call `npm install` in both of the subdirectories after removing them from the workspace, so I don't think I could just have it in the top level `package.json`.

.github/workflows/main-workflow.yml
alextaing: I think it makes sense to put it in there since it's an "automated fix". Moved!

.github/workflows/update-third-party-notices.yml
alextaing: The way that I reworked the process makes it so that both actions are encapsulated within one bash script, so I won't be able to use Turbo for parallelization, unfortunately.  I was trying to figure out a way to do this all in the `yml` workflow and avoid bash scripts, but setting runtime variables (which is something that I needed to be able to do) was not as straightforward as I thought.

generate-third-party-notices.sh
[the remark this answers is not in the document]
alextaing: I made a new approach where instead of moving the contents of `studio` and `studio-plugin` to a file outside of the workspace, I just temporarily remove `studio` and `studio-plugin` from the workspace, and re-add them once I finish making the third party notices.

package.json
```diff
@@ -27,11 +27,12 @@
     "eslint-plugin-react-perf": "^3.3.1",
     "eslint-plugin-tsdoc": "^0.2.17",
     "eslint-plugin-unused-imports": "^2.0.0",
+    "json": "^11.0.0",
```
alextaing, line 30: This is the[ lib that I use](https://github.com/trentm/json) to edit the package.json using cli.

.github/workflows/update-third-party-notices.yml
alextaing: I have a new approach where I don't think comment applies anymore, but LMK if you think it still does.  Mostly because I use a command that won't work correctly outside of GH actions since it requires the changing of workspaces to work correctly.

.github/workflows/update-third-party-notices.yml
oshi97: sure that's fair! can we have a more specific version then, instead of just v2

.prettierignore
oshi97, line 7: add newline

.github/workflows/update-third-party-notices.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      # this is done since studio is a monorepo.  a fresh reinstall without workspaces allows all
+      # node_modules to be installed within the current dir (instead of some at the root, with workspaces)
+      # NOTE: since only changes to THIRD-PARTY-NOTICES files are added, this change will not be pushed
+      - name: Clear workspace from package.json
+        run: json -I -f package.json -e 'this.workspaces=[]'
+
+      # npm install to generate node_modules which are required for generate-license-file to work
+      - working-directory: packages/studio
+        run: npm install; generate-license-file --input package.json --output THIRD-PARTY-NOTICES --overwrite
```
oshi97, line 31: why do you need to run npm install before?

.github/workflows/update-third-party-notices.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Update THIRD-PARTY-NOTICES
+        id: update-notices
+        uses: EndBug/add-and-commit@v9
+        with:
+          message: "Automated update to THIRD-PARTY-NOTICES from github action's 3rd party notices check"
+          add: '["packages/studio/THIRD-PARTY-NOTICES", "packages/studio-plugin/THIRD-PARTY-NOTICES"]'
+          push: true
+          default_author: github_actions
+      - name: Fail on push of updated third party notice
```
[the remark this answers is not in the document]
alextaing: oh true, that was leftover from when I had it in automated fixes-- thanks

.github/workflows/main-workflow.yml
```diff
@@ -7,13 +7,17 @@ concurrency:
 
 jobs:
   automated-fixes:
```
nmanu1, line 9: nit: rename to automated-linting

.github/workflows/update-third-party-notices.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  license-check:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
+          registry-url: "https://registry.npmjs.org"
+      - run: npm install -g json@v11.0.0
+      - run: npm install -g generate-license-file@2.0.0
```
nmanu1, line 17: nit: can we install both packages in the same command? npm install should accept multiple package names

PR #330, opened 2023-08-10
Components: Allow field picker to scroll with page

briantstephan:
This fixes an issue where the field picker dropdown would not scroll when the page scrolled. This also fixes an issue with the prop editor menu button appearing outside of the visible area.

J=SLAP-2859
TEST=auto,manual

Checked that the field picker dropdown scrolls appropriately and that the prop editor menu button appears next to the ellipses icon for all sidebar widths as expected. Also added a test for the menu button.


packages/studio/src/components/ActiveComponentPropEditors.tsx
```diff
@@ -32,13 +32,11 @@ export default function ActiveComponentPropEditors(props: {
 
   const propValues = activeComponentState.props;
   return (
-    <div className="overflow-x-auto">
```
nmanu1, line 35: was it necessary for getting the correct dropdown styling to move the overflow into the EditorSidebar? I preferred having the scrolling limited to ActiveComponentPropEditors because I think the spacing for the top toggle bar becomes a little weird if it's scrollable (see video)

https://github.com/yext/studio/assets/88398086/e0f99782-5922-45c7-8edc-35c2b5ecff1a

e2e-tests/__screenshots__/win32/undefined-menu-button.spec.ts/renders-undefined-menu-button-1.png
nmanu1, line 1: could we have the undefined menu appear a little more to the left of the button (so its right side is more closely aligned with button)? that way it would be less likely to require the user to scroll to see the full text

briantstephan: Done

packages/studio/src/components/ActiveComponentPropEditors.tsx
briantstephan: Fixed

packages/studio/src/components/ObjectPropEditor.tsx
```diff
@@ -45,7 +45,7 @@ export default function ObjectPropEditor(props: {
   return (
     <div className={containerClasses}>
       {renderBranchUI(isNested)}
-      <div>
+      <div className="min-w-[112px]">
```
alextaing, line 48: just out of curiosity-- what's the goal of this `min-w-[112px]`?

briantstephan: It ensures that the undefined menu button is always at least a minimum distance away from the left edge of the right sidebar. This prevents the button from getting partially cut off by going out of bounds when the prop editor would otherwise be very short.

oshi97 (approved):

packages/studio/src/components/ActiveComponentPropEditors.tsx
```diff
@@ -32,7 +32,7 @@ export default function ActiveComponentPropEditors(props: {
...
+    <div className="overflow-x-auto grow flex flex-col h-fill">
```
nmanu1, line 35: what does `h-fill` do? I'm only familiar with `h-full`

packages/studio/src/components/ObjectPropEditor.tsx
nmanu1: I think it looks a little weird to have the ellipses icon for the undefined menu appear a seemingly arbitrary distance to the right of short prop inputs, rather than immediately next to them. I'd rather have the menu be left-aligned with the ellipses icon than move the icon

packages/studio/src/components/PropsPanel.tsx
```diff
@@ -33,11 +33,13 @@ export default function PropsPanel(): JSX.Element | null {
           <Divider />
         </>
       )}
-      <ActiveComponentPropEditors
-        activeComponentState={extractedComponentState}
-        propShape={propShape}
-      />
-      <RepeaterPanel />
+      <div className="grow flex flex-col h-fill">
```
nmanu1, line 36: what are these classes adding? I see they're in a couple components

alextaing: not sure about the others, but I found that the `grow flex` makes sure that the dropdowns aren't cut off
image

packages/studio/src/components/UndefinedMenuButton.tsx
```diff
@@ -70,9 +70,9 @@ export default function UndefinedMenuButton({
           aria-label="Toggle undefined value menu"
         />
         {isOpen && (
-          <div className="absolute mt-3">
+          <div className="relative mt-3 left-[134px]">
             <ul
-              className="relative bg-white rounded border z-10 shadow-2xl text-sm"
+              className="absolute bg-white rounded border z-10 shadow-2xl text-sm"
```
alextaing, line 75: small thing, but I think with these changes, the dropdown gets cut off if it extends past the last prop editor
image

packages/studio/src/components/PropsPanel.tsx
briantstephan: Yes, it makes the component extend as necessary so that the dropdown isn't cut off. It needs to be in the dropdown's parent components as well so that they also grow, otherwise the child can only grow to the height of the parent.

packages/studio/src/components/ActiveComponentPropEditors.tsx
briantstephan: Meant to put `h-full` but it seems to not be necessary here anyway.

packages/studio/src/components/ObjectPropEditor.tsx
briantstephan: Done

packages/studio/src/components/UndefinedMenuButton.tsx
```diff
@@ -70,9 +70,9 @@ export default function UndefinedMenuButton({
           aria-label="Toggle undefined value menu"
         />
         {isOpen && (
-          <div className="absolute mt-3">
+          <div className="relative mt-3 left-[250px]">
```
nmanu1, line 73: I checked out the branch locally, and it looks like when the menu is open, the scrolling area extends much further than what's needed to display the menu. I think that's because of the `left-[250px]` here. If we remove this, can we tweak the positioning of the `ul` below to get the desired position without the extra overflow?

briantstephan: Good catch, I didn't notice that. Fixed.

packages/studio/src/components/UndefinedMenuButton.tsx
```diff
@@ -70,9 +70,9 @@ export default function UndefinedMenuButton({
           aria-label="Toggle undefined value menu"
         />
         {isOpen && (
-          <div className="relative mt-3 left-[250px]">
+          <div className="relative mt-3">
             <ul
-              className="absolute bg-white rounded border z-10 shadow-2xl text-sm"
+              className="absolute bg-white rounded border z-10 shadow-2xl text-sm left-[-13px]"
               style={listStyles}
```
nmanu1, line 76: nit: can we remove the `right` class from `listStyles` now that we specify `left` in the `className`?

e2e-tests/__screenshots__/darwin/undefined-menu-button.spec.ts/renders-undefined-menu-button-1.png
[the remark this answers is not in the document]
briantstephan: The right sidebar is horizontally scrollable, so the button should still always be accessible. The menu is now left aligned with the button to ensure that it never goes off the left edge of the sidebar, where it isn't scrollable (https://github.com/yext/studio/pull/330#discussion_r1293897094).

e2e-tests/tests/undefined-menu-button.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { expect } from "@playwright/test";
+import { studioTest } from "./infra/studioTest.js";
+
+studioTest("renders undefined menu button", async ({ page, studioPage }) => {
+  await studioPage.setActiveComponent("Banner");
+  await page.getByLabel("Toggle undefined value menu").first().click();
+  await expect(page).toHaveScreenshot();
```
nmanu1, line 7: can you add a screenshots to check the that if you scroll to the right, you're able to see the whole menu text? ideally we'd also check that if you scroll down, the menu remains next to the button, but that might be hard to implement with our current setup until we have more realistic test pages

nmanu1 (dismissed):
lgtm!

nmanu1 (approved):

PR #332, opened 2023-08-14
Fixes for starting the pages dev server

oshi97:
Updates the code to not kill everything after 100s.

J=NONE
TEST=manual

I can ctrl + c studio and after a 10s delay pages dev will end and studio will exit
I don't like this delay and debated between using a spawn vs spawnSync/execSync, but I think long term we just
need to start the pages dev server by importing createServer directly instead of via npx

### DIFF
--- a/packages/studio-plugin/src/createStudioPlugin.ts
+++ b/packages/studio-plugin/src/createStudioPlugin.ts
@@ -18,7 +18,7 @@ import { UserConfig } from "vite";
 import { STUDIO_PROCESS_ARGS_OBJ } from "./constants";
 import { startPagesDevelopmentServer } from "./startPagesDevelopmentServer";
 import LocalDataMappingManager from "./LocalDataMappingManager";
-import { execSync } from "child_process";
+import { execSync } from "node:child_process";
 
 /**
  * Handles server-client communication.


packages/studio-plugin/src/createStudioPlugin.ts
```diff
@@ -18,7 +18,7 @@ import { UserConfig } from "vite";
 import { STUDIO_PROCESS_ARGS_OBJ } from "./constants";
 import { startPagesDevelopmentServer } from "./startPagesDevelopmentServer";
 import LocalDataMappingManager from "./LocalDataMappingManager";
-import { execSync } from "node:child_process";
+import { execSync } from "child_process";
```
alextaing, line 21: I wasn't familiar with the `node:___` syntax, so I took a look into it--  I found out that it means that it makes sure that the package is from node core and not something else.  Just curious why you removed it

oshi97: oh hmm it's just not how we've been importing things
https://2ality.com/2021/12/node-protocol-imports.html
I'm definitely down to make an item to add an eslint rule for it!

alextaing (approved):
🔥

PR #333, opened 2023-08-14
Rework HMR to not use handleHotUpdate

oshi97:
This PR separates our custom HMR from vite's handleHotUpdate hook.
Instead, we now create our own watcher just for a studio user's src files. 
Overall, this should be more maintanable in the future, since vite's default HMR does not work well for a studio user's src files due to the atypical way we're using using them.

We also now fire HMR updates when a user updates the body of a component, which is behavior we had at one point but lost somewhere along the way. A test was added for this behavior.

The issue I was running into was that, after adding a custom tailwind config in a [separate PR](https://github.com/yext/studio/pull/323), whenever a page was deleted from the filesystem, studio would do a full page refresh. This was occurring due to tailwind registering the user's src files as dependencies 
(specifically right here https://github.com/tailwindlabs/tailwindcss/blob/ff2637257070c2ce8584603424d0063eba422b27/src/lib/setupTrackingContext.js#L126). 
This would cause the file to be hooked up to vite's 'unlnk' listener 
https://github.com/vitejs/vite/blob/77175099489ffbabefb9d98d448c782ced2986b5/packages/vite/src/node/server/index.ts#L560
The behavior of the unlink listener skips the handleHotUpdate hook and cannot be customized. When this same issue popped up in https://github.com/yext/studio/pull/304 for the localData folder, we could simply add the localData folder to the watcher's ignore options. We can't do that for a user's page templates, because completely ignoring them would mean we can't perform HMR updates on them.

The new architecture imo is simpler to understand because it only deals with studio specific details instead of dealing with all vite hot updates (which we don't need to touch since they work just the way we'd like and we're not doing anything weird outside of how we use the user's src files).

J=SLAP-2883,SLAP-2607
TEST=manual,auto

tested hot module reloading works for editing a page and editing a component
I can delete a page via UI and hit save and there is no full page reload
tests for adding and removing new components and pages


packages/studio-plugin/src/HmrManager.ts
```diff
@@ -1,78 +1,65 @@
...
-   * Currently we ignore all files under localData, to avoid a full page refresh when
-   * the generate-test-data yext CLI command is called
-   */
-  shouldExcludeFromWatch = (filepath: string): boolean => {
-    return upath.normalize(filepath).startsWith(this.userPaths.localData);
-  };
-
-  private static async reloadAssociatedModules(ctx: HmrContext) {
-    const reloadModulePromises = ctx.modules.map((m) => {
-      return ctx.server.reloadModule(m);
```
nmanu1, line 52: what was this needed for before?

packages/studio-plugin/src/HmrManager.ts
```diff
@@ -1,78 +1,65 @@
...
+    private server: ViteDevServer,
     private orchestrator: ParsingOrchestrator,
-    private pathToUserProjectRoot: string,
     private userPaths: UserPaths
   ) {}
 
+  createWatcher(pathToUserProjectRoot: string) {
+    const watcher = chokidar.watch(upath.join(pathToUserProjectRoot, "src"));
+    watcher.on("change", this.handleHotUpdate);
+    watcher.on("unlink", this.handleHotUpdate);
```
nmanu1, line 21: would we also want to watch for new files added?

packages/studio-plugin/src/viteconfig/getStudioViteOptions.ts
```diff
@@ -0,0 +1,23 @@
...
+  pathToUserProjectRoot: string
+): UserConfig {
+  const serverOptions: ServerOptions = {
+    port: studioConfig.port,
+    open:
+      args.mode === "development" &&
+      args.command === "serve" &&
+      studioConfig.openBrowser,
+    watch: {
+      ignored: pathToUserProjectRoot,
```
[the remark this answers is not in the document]
oshi97: I think so, they get ignored here though

packages/studio-plugin/src/HmrManager.ts
oshi97: This was emulating default hmr behavior
I think it might have been possible before to just return undefined if we wanted default behavior

packages/studio-plugin/src/HmrManager.ts
oshi97: This wasn't something we had before and was hard to do before (https://github.com/yext/studio/pull/123) but with some tweaks I added it here! also added tests

oshi97:
bumped playwright timeouts because studio startup was still timing out in ci occasionally. I believe this is partially due to the increased resource usage from running `pages dev` on startup

packages/studio-plugin/src/HmrManager.ts
nmanu1: would we still want default hmr behavior for files in the project root that Studio doesn't care about?

nmanu1: I'm confused, I thought `reloadAssociatedModules` was the default HMR behavior that we had for all files (except `localData` files) and we just had extra behavior if the file was in the project root. Now, HMR for files outside the project root is handled by Vite and HMR for files inside the `src` directory in the project root is handled by `handleHotUpdate`. But I think files in the project root, but not in the `src` directory, wouldn't have any HMR handling now, when we used to reload the modules before?

oshi97: it would be called before but wouldnt functionally do anything because no other modules would be registered as importers

oshi97: We also want to ignore localData and just generally speaking vite's HMR will not work properly with any of the user's own files

oshi97: I might be missing something but I'm not sure how that makes it more clear there's only two cases?

nmanu1: ah I forgot about localData. makes sense as is then!

nmanu1 (approved):

PR #335, opened 2023-08-14
Include Static Pages in `GenerateTestData`

alextaing:
In https://github.com/yext/studio/pull/304, we generate test data for new entity pages when `GenerateTestData` is called.  This PR changes this to include new static pages as well. 

J=SLAP-2866
TEST=auto, manual

It was checked that when a new static page is created, it is also added in the `mapping.json` file generated by `GenerateTestData`.  Tests were modified to account for the new change that creation of static pages also call `GenerateTestData`.


oshi97 (approved):
banzai

PR #336, opened 2023-08-15
Components: Mobile Viewport

asanehisa:
Right now there are an assortment of preset options for viewports. An issue I ran into was [tailwind not accepting dynamic class names]( https://tailwindcss.com/docs/content-configuration#dynamic-class-names) so I had to manually preset the aspect ratios. Will add more tests soon. 

I can also change the icon + location, I just took it from [brandfolder](https://brandfolder.com/yext/#!asset/q4zct2-2u6azc-9dlblt)

J=SLAP-2839
TEST=auto, manual

[loom demo. ](https://www.loom.com/share/6ff6595ede5b407ba4555b9889149cb8)


asanehisa:
Todo: update playwright
Playwright failing due to screenshot diff. It's b/c the viewport button is in action bar now

packages/studio/src/App.tsx
```diff
@@ -3,17 +3,29 @@ import ActionsBar from "./components/ActionsBar";
 import Toast from "./components/Toast";
 import PreviewWithUseComponents from "./components/PreviewWithUseComponents";
 import LeftSidebar from "./components/LeftSidebar";
+import { ViewportStyles } from "./components/Viewport/defaults";
+import { useRef, useState } from "react";
 
 export default function App() {
+  const previewRef = useRef<HTMLDivElement>(null);
```
tmeyer2115, line 10: Instead of adding the viewport state here, at the top-level of the app, I think it could be beneficial to put this in the Studio Store. I'm thinking we create a `PagePreview` slice and add that to the Store. Right now, it would just contain viewport info. Later, we could migrate things like `setActiveEntityFile` and `setActivePage` over to this Slice as well. It'd be the source of truth for everything Preview related. 

I think this should ultimately simplify the code. We don't need to worry about passing these new props to `PreviewWithUseComponents` or `ActionsBar`. Instead, internally, Components could access/set the viewport with our `useStudioStore` hook.

oshi97:
it's a bit of an idiosyncrasy, but after windows snapshots are committed via CI we generally add an empty commit to trigger the CI checks again

packages/studio/src/App.tsx
asanehisa: done I think. I put viewportDimensions + setViewportDimensions in the slice

packages/studio/src/components/PreviewWithUseComponents.tsx
```diff
@@ -19,11 +25,23 @@ export default function PreviewWithUseComponents() {
...
+    viewportDimensions.height * (previewRef.current?.clientWidth ?? 0)
+      ? " w-full "
+      : " h-full ";
+  const id = viewportDimensions.name.replace(/\s+/g, "").toLowerCase();
+  const iframeCSS = twMerge(
+    "mr-auto ml-auto border-2",
+    EXPANDED_VIEWPORTS[id].css,
+    dimensions
+  );
+
```
oshi97, line 39: I was thinking we could move this logic and also the `<div ref={previewRef} className="grow w-1/3 bg-white border-8 shadow">` div into IFramePortal?

packages/studio/src/components/Viewport/ViewportButton.tsx
```diff
@@ -0,0 +1,29 @@
+import { useCallback, useRef, useState } from "react";
+import useRootClose from "@restart/ui/useRootClose";
+import ViewportMenu from "./ViewportMenu";
+
+export default function ViewportButton(): JSX.Element {
+  const [isOpen, setIsOpen] = useState(false);
+  const containerRef = useRef<HTMLDivElement>(null);
+  const icon =
+    "https://assets.brandfolder.com/q4zcv4-5x9x3c-cwpfeb/v/9458449/original/DevicesLaptopPhoneTablet_B.png";
```
oshi97, line 9: Is there an svg version of this image that we can commit to the repo? Then we could directly import the svg as a react component (we do this in a few places like `import { ReactComponent as Globe } from "../icons/globe.svg";`)

packages/studio/src/components/Viewport/ViewportMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+  );
+}
+
+function Selector(props: { afterSelect: () => void }) {
+  const { afterSelect } = props;
+
+  return (
+    <div className="flex flex-col items-start py-1">
+      {Object.values(EXPANDED_VIEWPORTS).map((val) => {
+        return <Option afterSelect={afterSelect} {...val} />;
```
oshi97, line 27: missing key prop!

packages/studio/src/components/Viewport/ViewportButton.tsx
tmeyer2115: (Or we could convert the png to an svg if there isn't a similar thing for png files)

packages/studio/src/components/Viewport/ViewportMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const { closeMenu } = props;
+
+  return (
+    <div className="absolute z-20 rounded bg-white text-sm text-gray-700 shadow-lg">
+      <Selector afterSelect={closeMenu} />
+    </div>
+  );
+}
+
+function Selector(props: { afterSelect: () => void }) {
```
oshi97, line 21: might be a little simpler to not have a separate Selector component

packages/studio/src/components/Viewport/ViewportMenu.tsx
```diff
@@ -0,0 +1,67 @@
+import { useCallback } from "react";
+import { EXPANDED_VIEWPORTS, Viewport } from "./defaults";
+import useStudioStore from "../../store/useStudioStore";
+
+interface optionProps extends Viewport {
```
oshi97, line 5: nit capital O

packages/studio/src/components/Viewport/ViewportMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    <button
+      className="flex items-center gap-x-2 px-6 py-2 cursor-pointer hover:bg-gray-100 disabled:opacity-25 w-full text-left"
+      onClick={handleSelect}
+      aria-label={`Add ${viewportOption} Element`}
+    >
+      <div className="flex flex-row gap-x-2 items-center">
+        {viewportOption}
+        <div className="opacity-75">
+          {props.styles ? (
```
oshi97, line 58: can use `&&` here instead of `?` plus fragment

packages/studio/src/components/Viewport/ViewportMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+function Option(props: optionProps) {
+  const [setViewportDimensions] = useStudioStore((store) => [
+    store.pagePreview.setViewportDimensions,
+  ]);
+  const { afterSelect } = props;
+  const viewportOption = props.name;
+
+  const handleSelect = useCallback(() => {
+    afterSelect?.();
```
oshi97, line 41: afterSelect is required in `optionProps` so you can get rid of the optional chaining

packages/studio/tests/components/ViewportButton.test.jsx
```diff
@@ -0,0 +1,7 @@
+import { render, screen } from "@testing-library/react";
+import ViewportButton from "../../src/components/Viewport/ViewportButton";
+
+it("renders viewport button", () => {
+  render(<ViewportButton />);
```
oshi97, line 5: it's actually possible to test that clicking things in the button affects the store! You can do something like 
```ts
expect(useStudioStore.getState().blah.state.whatever).toEqual(something);
```

packages/studio/src/components/Viewport/defaults.tsx
```diff
@@ -0,0 +1,124 @@
...
+    styles: {
+      height: 720,
+      width: 600,
+    },
+    type: "mobile",
+    css: "aspect-[600/720]",
+  },
+};
+
+export const MINIMAL_VIEWPORTS: ViewportMap = {
```
oshi97, line 90: This value seems unused? I don't have a strong opinion for whether we use EXPANDED_VIEWPORTS or MINIMAL_VIEWPORTS

asanehisa: Yup unused rn, left it so if we wanna switch it's easy. I can ask Tom

asanehisa: Screenshot 2023-08-16 at 4 18 21 PM
Screenshot 2023-08-16 at 4 19 23 PM

asanehisa: Tom said expanded!

packages/studio/src/components/IFramePortal.tsx
```diff
@@ -1,27 +1,50 @@
...
   }>
 ) {
+  const previewRef = useRef<HTMLDivElement>(null);
   const iframeDocument = props.iframeEl?.contentWindow?.document;
+  const [viewportDimensions] = useStudioStore((store) => [
+    store.pagePreview.viewportDimensions,
+  ]);
+  const dimensions =
+    viewportDimensions.width * (previewRef.current?.clientHeight ?? 0) >
+    viewportDimensions.height * (previewRef.current?.clientWidth ?? 0)
```
tmeyer2115, line 27: For my own edification, could you give me a bit more detail on this logic? At a high-level I want to understand when we need to switch between w-full and h-full.

packages/studio/src/components/Viewport/defaults.tsx
```diff
@@ -0,0 +1,88 @@
...
+  name: string;
+  height: number;
+  width: number;
+}
+
+export interface ViewportMap {
+  [key: string]: Viewport;
+}
+
+export const EXPANDED_VIEWPORTS: ViewportMap = {
```
tmeyer2115, line 18: Small nit: Could we rename `EXPANDED_VIEWPORTS`? I think the term "Expanded" doesn't make too much sense now that we've removed "Minimal".

packages/studio/src/components/Viewport/ViewportMenu.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <div className="absolute z-20 rounded bg-white text-sm text-gray-700 shadow-lg flex flex-col items-start py-1">
+      {Object.values(EXPANDED_VIEWPORTS).map((val) => {
+        return <Option key={val.name} closeMenu={props.closeMenu} {...val} />;
+      })}
+    </div>
+  );
+}
+
+function Option(props: OptionProps) {
```
tmeyer2115, line 21: Small nit, could we do `Option({ name, closeMenu, styles }: OptionProps)` here? It'd simplify some of the code below.

packages/studio/src/components/Viewport/ViewportMenu.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const [setViewportDimensions] = useStudioStore((store) => [
+    store.pagePreview.setViewportDimensions,
+  ]);
+  const { closeMenu } = props;
+  const viewportOption = props.name;
+
+  const handleSelect = useCallback(() => {
+    closeMenu();
+    setViewportDimensions({
+      name: props.name ?? "Reset Viewport",
```
tmeyer2115, line 31: Since `OptionProps` extends `ViewPort`, won't `props.name` always be defined? It looks like all the entries in `EXPANDED_VIEWPORTS` will have a non-empty string for name.

packages/studio/src/components/IFramePortal.tsx
```diff
@@ -1,27 +1,50 @@
...
+      ? " w-full "
+      : " h-full ";
+  const id = viewportDimensions.name.replace(/\s+/g, "").toLowerCase();
+  const iframeCSS = twMerge(
+    "mr-auto ml-auto border-2",
+    EXPANDED_VIEWPORTS[id].css,
```
tmeyer2115, line 33: What do you think of also storing the current viewport's CSS in `store.pagePreview`? Then, we wouldn't need to do this lookup.

packages/studio/src/components/Viewport/ViewportMenu.tsx
```diff
@@ -0,0 +1,51 @@
...
+      height: props.styles?.height ?? window.innerHeight,
+      width: props.styles?.width ?? window.innerWidth,
+    });
+  }, [closeMenu, setViewportDimensions, props]);
+
+  return (
+    <button
+      className="flex items-center gap-x-2 px-6 py-2 cursor-pointer hover:bg-gray-100 disabled:opacity-25 w-full text-left"
+      onClick={handleSelect}
+      aria-label={`Add ${viewportOption} Element`}
```
alextaing, line 41: Would this be more like `Select ${viewportOption} Viewport` ?

packages/studio/src/components/Viewport/ViewportMenu.tsx
```diff
@@ -0,0 +1,51 @@
...
+    setViewportDimensions({
+      name: props.name ?? "Reset Viewport",
+      height: props.styles?.height ?? window.innerHeight,
+      width: props.styles?.width ?? window.innerWidth,
+    });
+  }, [closeMenu, setViewportDimensions, props]);
+
+  return (
+    <button
+      className="flex items-center gap-x-2 px-6 py-2 cursor-pointer hover:bg-gray-100 disabled:opacity-25 w-full text-left"
```
alextaing, line 39: Are there situations where we need `disabled`?

alextaing:
Should we write a playwright test for this? (I vaguely remember someone mentioning we might be holding off on some of those?)

alextaing:
I noticed that if you select a mobile viewport then resize the window, it leaks out into the prop editor panel if the window is too small.  Is this something we want to fix in this item?

image

packages/studio/src/components/IFramePortal.tsx
```diff
@@ -1,27 +1,50 @@
...
-import { Dispatch, PropsWithChildren, SetStateAction, useEffect } from "react";
+import {
+  Dispatch,
+  PropsWithChildren,
+  SetStateAction,
+  useEffect,
+  useRef,
+} from "react";
+import useStudioStore from "../store/useStudioStore";
+import { EXPANDED_VIEWPORTS } from "./Viewport/defaults";
```
alextaing, line 10: I noticed that if you select a mobile viewport then resize the window, it leaks out into the prop editor panel if the window is too small.  Is this something we want to fix in this item?

image

alextaing: There's also interesting behavior where if the window is small, it seems like the viewport will grow if the window is made larger-- however if the window is large and is made smaller, the viewport won't shrink. Also, when the viewport resizes, it also appears that the contents of the viewport remain the same size, which I think is unexpected. I think that all these quirks are just caused by window resizing.

packages/studio/src/components/IFramePortal.tsx
asanehisa: So it's basically: if the ratio (width / height) of the viewport is greater than the ratio of the preview div then it should use w-full. If it instead used h-full, it would overflow from the div. 
This is because the aspect ratio works best if it has some style letting it know if height or width have pre-set dimensions. 

I rearranged the func to use * instead of / though.

packages/studio/src/components/IFramePortal.tsx
asanehisa: Hmmm I'm not too sure. It doesn't make total sense to me rn, right now it's calling the newly clicked viewport's css (which I hardcoded in Viewport/defaults). I could hardcode them into the store.pagePreview instead? As a key: value graph...?

packages/studio/src/components/Viewport/ViewportMenu.tsx
asanehisa: Oops ur very right

packages/studio/src/components/Viewport/ViewportMenu.tsx
asanehisa: nope! removing!

packages/studio/src/store/models/slices/PagePreviewSlice.ts
```diff
@@ -0,0 +1,6 @@
+import { ViewportStyles } from "../../../components/Viewport/defaults";
+
+export default interface PagePreviewSlice {
+  viewportDimensions: ViewportStyles;
```
tmeyer2115, line 4: What I'm thinking is we have this:

```
viewport: Viewport;
setViewport(viewport: Viewport);
```

In the drop-down you've created, we'd modify `handleSelect` to look like this:

```
  const handleSelect = useCallback(() => {
    closeMenu();
    const viewport = VIEWPORT.values().findFirst(viewport.name === name);
    // Error handling to make sure `viewport` exists.
    // Apply default styles to viewport if `styles.height` and `styles.width` are missing
    setViewport(viewport);
  }, [closeMenu, name, setViewportDimensions, styles]);
```

In my mind, the main benefit to this approach is that we don't need to look in two different places for information about the Viewport. Right now, in `IFramePortal`, we get the dimensions of the current Viewport from global state, but the CSS from the static map. It'd be nice if we could get the entire, current `Viewport` from State. If this doesn't work because of Tailwind dynamic class name issues, lemme know!

packages/studio/src/components/IFramePortal.tsx
```diff
@@ -1,27 +1,59 @@
...
+      (viewport.styles?.height ?? 0) * (previewRef.current?.clientWidth ?? 0)
+        ? " w-full "
+        : " h-full ";
+    if (tempCss !== css) {
+      setCss(tempCss);
+    }
+  };
+
+  useEffect(() => {
+    window.addEventListener("resize", handleResize);
```
oshi97, line 43: is it possible to use a ResizeObserver here?

packages/studio/src/components/IFramePortal.tsx
```diff
@@ -1,27 +1,59 @@
...
   props: PropsWithChildren<{
-    className?: string;
     title: string;
     iframeEl: HTMLIFrameElement | null;
     setIframeEl: Dispatch<SetStateAction<HTMLIFrameElement | null>>;
   }>
 ) {
+  const previewRef = useRef<HTMLDivElement>(null);
   const iframeDocument = props.iframeEl?.contentWindow?.document;
+  const [viewport] = useStudioStore((store) => [store.pagePreview.viewport]);
```
oshi97, line 22: Could we move all the viewport css logic into a hook at the bottom of the file?

packages/studio/src/components/PreviewWithUseComponents.tsx
```diff
@@ -7,23 +7,22 @@ import { Tooltip, ITooltip } from "react-tooltip";
...
-    store.actions.getComponentTree()
-  );
   const [tooltipProps, setTooltipProps] = useState<ITooltip>({
     isOpen: false,
     position: { x: 0, y: 0 },
     content: "",
     anchorSelect: "",
   });
   const [iframeEl, setIframeEl] = useState<HTMLIFrameElement | null>(null);
+  const [componentTree] = useStudioStore((store) => [
```
oshi97, line 17: nit this doesn't need to be in a tuple

packages/studio/src/components/Viewport/ViewportMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+      })}
+    </div>
+  );
+}
+
+function Option({ name, closeMenu, styles }: OptionProps) {
+  const [setViewport] = useStudioStore((store) => [
+    store.pagePreview.setViewport,
+  ]);
+  const viewportOption = name;
```
oshi97, line 25: nit: you can do the rename above like `function Option({ name: viewportOption, which prevents the extra variable. Below the `name` variable is used in the callback but `viewportOption` is used in the markup when we probably just want to use one variable name

packages/studio/src/components/IFramePortal.tsx
asanehisa: I think so? Would that be better?

oshi97: It should be! resize event listeners are notoriously laggy (here's a random article I googled there's probably better ones tbh https://tigeroakes.com/posts/resize-event-alternatives/)

packages/studio/src/components/IFramePortal.tsx
```diff
@@ -70,3 +55,29 @@ function useParentDocumentStyles(iframeDocument: Document | undefined) {
     }
   }, [iframeDocument]);
 }
+
+const GetCSS = (viewport: Viewport, previewRef: RefObject<HTMLDivElement>) => {
```
tmeyer2115, line 59: Could we call this `useCSS`? With React Hooks, there's a general naming convention for the hook's name to start with `use`.

packages/studio/src/components/IFramePortal.tsx
asanehisa: oooh got it!

tmeyer2115 (dismissed):

alextaing (approved):

PR #338, opened 2023-08-16
Fix xdg-open error in CBD

oshi97:
Vite uses xdg-open in order to open a browser window on start up.
xdg-open is not guaranteed to exist, and actually doesn't exist in the CBD images.
This causes errors like the below
image

It doesn't make sense to open a browser in CBD anyways so I set the flag to false for this scenario.

J=SLAP-2886
TEST=manual

I can run studio in CBD without error

### DIFF
--- a/packages/studio-plugin/src/viteconfig/getStudioViteOptions.ts
+++ b/packages/studio-plugin/src/viteconfig/getStudioViteOptions.ts
@@ -11,6 +11,7 @@ export default function getStudioViteOptions(
     open:
       args.mode === "development" &&
       args.command === "serve" &&
+      !process.env.YEXT_CBD_BRANCH &&
       studioConfig.openBrowser,
     watch: {
       ignored: pathToUserProjectRoot,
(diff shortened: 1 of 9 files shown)


packages/studio-plugin/src/viteconfig/getStudioViteOptions.ts
```diff
@@ -11,6 +11,7 @@ export default function getStudioViteOptions(
     open:
       args.mode === "development" &&
       args.command === "serve" &&
+      !process.env.YEXT_CBD_BRANCH &&
```
nmanu1, line 14: would it be better to check that `xdg-open` exists instead of tying the check to CBD details?

oshi97: I can't find any way to check if it exists

nmanu1 (approved):

jwartofsky-yext (dismissed):
lgtm

PR #341, opened 2023-08-18
Update to support ^1.0.0-rc.1

jwartofsky-yext:
Update to support ^1.0.0-rc.0

J=SLAP-2892
TEST=auto,manual

Auto tests pass.

Can run studio with npm run dev using
^1.0.0-rc.0 after modification.

Note that this change requires [this](https://github.com/yext/pages/commit/4cf67c99207883500b213e8aa68d9c6704c0210e) update in PagesJS, so will either need to update PagesJS locally or wait until the next PagesJS release.


jwartofsky-yext:
This should not be merged until the next version of PagesJS is released to npm. This PR will also need to be updated to use the new PagesJS version.

jwartofsky-yext:
Tested with dyext/pages-1.0.0-rc.1 and npm run dev. All is working as expected. This should be good to go.

e2e-tests/vite.config.js
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from "vite";
```
oshi97, line 1: remove this

apps/test-site/vite.config.js
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from "vite";
```
oshi97, line 1: Is it possible for pagesjs to provide a default vite config if the user doesn't have one? I'd rather not have extra vite configs floating around the studio repo when we're just using the recommended default.

oshi97: If it's quick could we make the change in PagesJS? We don't need to release it right away but I'd like to remove these at a later date

package.json
```diff
@@ -21,6 +21,8 @@
     "@typescript-eslint/parser": "^5.59.0",
     "@yext/eslint-config": "^1.0.0",
     "@yext/eslint-plugin-export-star": "^1.2.0",
+    "@yext/pages": "^1.0.0-rc.1",
```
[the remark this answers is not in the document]
jwartofsky-yext: I don't believe so. However this resolved the issue where we were pulling [this](https://www.npmjs.com/package/pages) package.

e2e-tests/tests/open-live-preview.spec.ts
```diff
@@ -9,6 +9,6 @@ studioTest(
     await studioPage.livePreviewButton.click({ position: { x: 75, y: 0 } });
     const popup = await popupPromise;
     await popup.waitForLoadState();
-    await expect(popup).toHaveTitle("Pages Development Page");
+    await expect(popup).toHaveTitle("Pages Development");
```
jwartofsky-yext, line 12: This change was needed to make the test pass

oshi97 (approved):

packages/studio-plugin/package.json
```diff
@@ -33,7 +33,7 @@
     "@babel/preset-env": "^7.22.4",
     "@babel/preset-react": "^7.22.3",
     "@babel/preset-typescript": "^7.21.5",
-    "@yext/pages": "^1.0.0-beta.14",
+    "@yext/pages": "^1.0.0-rc.1",
```
nmanu1, line 36: I think the third party notices for `studio-plugin` need to be updated to reflect this change

PR #342, opened 2023-08-21
Update Playwright test for rearranging elements

nmanu1:
This PR updates the Playwright test for rearranging elements so that it uses the new LocationPage.

J=SLAP-2888
TEST=auto


e2e-tests/tests/rearrange-elements.spec.ts
```diff
@@ -8,22 +8,31 @@ const expectedPage = fs.readFileSync(
...
+    (await page.mouse.move(
+      businessInfoBox.x,
+      businessInfoBox.y + businessInfoBox.height - 5,
+      {
+        steps: 1000,
+      }
+    ));
   await page.waitForTimeout(1000);
   await page.mouse.up();
   await expect(page).toHaveScreenshot();
```
oshi97, line 32: It might make sense to take screenshots for the component tree and preview instead of the whole page? I was thinking for here and on line 13

e2e-tests/tests/rearrange-elements.spec.ts
```diff
@@ -8,22 +8,31 @@ const expectedPage = fs.readFileSync(
 );
 
 studioTest("can rearrange elements in tree", async ({ page, studioPage }) => {
-  const banner = page.getByText("Banner");
-  const div = page.getByRole("list").filter({ hasText: "div" });
-  const divBox = await div.boundingBox();
+  await studioPage.switchPage("LocationPage");
+  await studioPage.addElement("Footer", "Components", false);
```
oshi97, line 12: Do we need to add a footer?

e2e-tests/tests/rearrange-elements.spec.ts
nmanu1: In the second vs. third screenshots, the main difference is the state of the Save button, so we'd want to capture the whole screen to compare them properly. And then I thought we might as well capture the whole screen for the first screenshot too to make the comparison easier between the first and second screenshots

e2e-tests/tests/rearrange-elements.spec.ts
oshi97: Oh ok hmm I think I prefer doing that instead? Then we can skip the first screenshot

e2e-tests/tests/rearrange-elements.spec.ts
oshi97: Could we take a screenshot of just the save + deploy button? We could also just check that the two buttons both have the right enabled/disabled states. I think right now it's possible that the screenshot is large enough where even if the state of the buttons did change not enough pixels would change for the test to fail

e2e-tests/tests/rearrange-elements.spec.ts
oshi97: I've been thinking we should avoid large full page screenshots where possible because they can result in false positives and we've also run into a number of "re-screenshot the world" scenarios when somebody adds a new UI element to the page

oshi97 (approved):

PR #344, opened 2023-08-21
Field Picker for Number-Type Fields

asanehisa:
This generates a field dropdown for number type fields. Instead of having it as [[businessId]] in the propbox it gets the appropriate number (i.e. 3350634) and has that in the prop box. Also PopulatePropInput is almost the same as StringPropInput, but with a different name (I can change this) and accepts numbers as well. 

J=SLAP-2881
TEST=auto, manual 

[loom](https://www.loom.com/share/a939e94b04114f9e8f28657a7d4d3ca0?sid=78736e87-28f1-41d2-bf48-cfbc266acaaa)


packages/studio/src/components/PropInput.tsx
```diff
@@ -61,17 +58,16 @@ export default function PropInput({
   switch (type) {
     case PropValueType.number:
       return (
-        <input
-          type="number"
-          onChange={handleChangeEvent}
-          className={inputBoxCssClasses}
-          value={(propValue ?? "") as number | string}
+        <PopulatePropInput
```
oshi97, line 61: I think it might make more sense to have a separate NumberPropInput component? Instead of using the same component for strings and numbers

asanehisa: I could! I was debating, but the code between the two would be pretty similar.

oshi97: Ideally the component for string props wouldn't have any code only used for number props and vice versa. You could still share code between the two of them via a shared component or hook. Alternatively you could turn the logic that is different between the two use cases as props instead of having switching logic within a component. At that point you'd probably just want two separate components though

packages/studio/src/components/NumberPropInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+      const newPropValue = get(
+        { document: expressionSources["document"] },
+        fieldId
+      );
+      onChange(newPropValue);
+    },
+    [onChange, expressionSources]
+  );
+
+  const fieldPickerNumberFilter = useCallback(
```
oshi97, line 43: This callback doesn't take in any dependencies so we can put it outside the body of the component like useExpressionSources. We can also update the corresponding one in StringPropInput

packages/studio/src/components/NumberPropInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+      const val = Number(e.target.value);
+      onChange(val);
+    },
+    [onChange]
+  );
+
+  const appendField = useCallback(
+    (fieldId: string) => {
+      const newPropValue = get(
+        { document: expressionSources["document"] },
```
oshi97, line 35: Right now if you use the field picker for say businessId, it'll add the number literal to the prop input like `12345`
However this isn't what we want since when this gets saved the component ends up looking like
```tsx
<MyComponent num={12345}/>
```
instead of
```tsx
<MyComponent num={document.businessId}/>
```

In studio we have a concept of Expressions and Literals via PropValueKind.Expression and PropValueKind.Literal. We need NumberPropInput to be able to handle both kinds. For example maybe the initial value for a number prop is a literal, but when they use the field picker it instead changes it into an expression

asanehisa: new [loom](https://www.loom.com/share/038cfc9931fe43dfa40d4b9fcb1d1256?sid=09348a93-b461-457d-a3ca-f25ba186218f). 

Hmmm so I have it working for expressions, but I kind of manually set it as an expression? Not my favorite way to fix it unfortunately, but I couldn't seem to find an earlier / easier way to change the PropValueKind of that component. It does however work + doesn't accept arbitrary strings (only allows strings when set via field dropdown)

packages/studio/src/components/NumberPropInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+    (e: ChangeEvent<HTMLInputElement>) => {
+      const val = Number(e.target.value);
+      onChange(val);
+    },
+    [onChange]
+  );
+
+  const appendField = useCallback(
+    (fieldId: string) => {
+      const documentUsage = "${" + fieldId + "}";
```
oshi97, line 32: This syntax only exists to support template string interpolation, and shouldn't be here for numbers

asanehisa: Oh? I was using it to help handle the expressions since we want it to accept stuff like document.businessId.

oshi97: you can just have it be `document.businessId` instead of `${document.businessId}`
when you save to file with the template string syntax what does it look like?

asanehisa: oh ur right! Ohhh I get why its string thing now

packages/studio/src/components/StringPropInput.tsx
```diff
@@ -47,19 +47,21 @@ export default function StringPropInput({
...
       displayValue={displayValue}
-      fieldFilter={fieldPickerFilter}
+      fieldFilter={useFieldPickerFilter()}
       disabled={disabled}
     />
   );
 }
+
+function useFieldPickerFilter() {
+  return useCallback(
```
alextaing, line 62: since this function is outside the component, it doesn't need the useCallback (correct me if im wrong!)

asanehisa: oooh u right

packages/studio/src/components/NumberPropInput.tsx
```diff
@@ -0,0 +1,48 @@
...
+    (fieldId: string) => {
+      onChange(fieldId);
+    },
+    [onChange]
+  );
+
+  return (
+    <FieldPickerInput
+      onInputChange={handleChangeEvent}
+      handleFieldSelection={appendField}
```
nmanu1, line 38: nit: can `onChange` be passed directly here instead of creating a new `appendField` function?

asanehisa: Hmmm yes, but it looks messier to me? 
Screenshot 2023-08-23 at 9 41 15 AM

nmanu1: oh I meant something like this:
```tsx
<FieldPickerInput
  // ... other props
  handleFieldSelection={onChange}
/>
```

asanehisa: Oh like removing the useCallback? When I do I get an autofix warning `JSX attribute values should not contain functions created in the same scope`

nmanu1: you get that error if you use the code snippet above? I'm pretty sure that's how we pass it in ArrayPropEditor

asanehisa: If I use this `handleFieldSelection={fieldId => onChange(fieldId)}`

nmanu1: ah that makes sense since that's creating a new function in line. you should be able to directly pass the function (`handleFieldSelection={onChange}`) since `onChange` is passed in as a prop to the component, rather than created in the body of the component

asanehisa: Ohhh I see! Thank you!

packages/studio/src/components/NumberPropInput.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * corresponding prop's value for the active component based on a user's input.
+ */
+export default function NumberPropInput({
+  value,
+  onChange,
+  disabled,
+}: PropInputProps): JSX.Element {
+  const handleChangeEvent = useCallback(
+    (e: ChangeEvent<HTMLInputElement>) => {
+      const val = Number(e.target.value);
```
oshi97, line 22: If val gets parsed to NaN can we prevent onChange from being called?

oshi97: yup exactly!

asanehisa: Done! Updated [loom](https://www.loom.com/share/a939e94b04114f9e8f28657a7d4d3ca0?sid=78736e87-28f1-41d2-bf48-cfbc266acaaa) since no more NaN

packages/studio/src/components/NumberPropInput.tsx
```diff
@@ -0,0 +1,43 @@
...
+      if (!isNaN(val)) {
```
oshi97, line 23: I don't think it really matters here since you first coerce it into a number, but I read online that `Number.isNaN` is more reliable than the global `isNaN`, so I was thinking we might as well use `Number.isNaN` instead

asanehisa: Oh very interesting! Done!

oshi97 (approved):
🔥 🔥 🔥

PR #346, opened 2023-08-22
Add version number

alextaing:
Product requested the current version number to be displayed at the top of the page to the left of the live preview button.

J=SLAP-2899
TEST=manual


packages/studio/src/components/VersionNumber.tsx
```diff
@@ -0,0 +1,8 @@
+import { version } from "../../package.json";
+
+/**
+ * Renders the current version number
+ */
+export default function VersionNumber() {
+  return <p className="text-gray-500">v.{version}</p>;
```
oshi97, line 7: can we also get the studio-plugin version?

alextaing: will implement in future! not included in this PR

alextaing:
Is it OK that none of the screenshots updated, even though there was a visual change?

oshi97:
> Is it OK that none of the screenshots updated, even though there was a visual change? I see that the version number does appear in the videos that Playwright takes.

we're getting false positives because the screenshots we're taking are too large, and the pixelmatch ratio isn't low enough to register a change
on the other hand we can't make the ratio too low, though, due to random jitter causing false negatives even if the snapshots are the "same"
going forward we want to avoid full page screenshots where possible and opt for more localized ones

on a side note this version number is going to be a little annoying moving forward because every time we update the version we'll have to rescreenshot the world for all full page screenshots
Maybe you could ask product if we could hide this version number behind a little icon or something you would hover over?

packages/studio/src/components/VersionNumber.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <p className="font-bold mb-4"> Studio Info</p>
+          Studio Verison: {version}
+        </ReactModal>
+      );
+    },
+    []
+  );
+
+  return (
+    <>
```
oshi97, line 50: extra fragment

packages/studio/src/components/VersionNumber.tsx
```diff
@@ -0,0 +1,58 @@
...
+  overlay: {
+    zIndex: 1000,
+  },
+};
+
+/**
+ * Renders the current version number
+ */
+export default function VersionNumber() {
+  const infoSVG = <Info className="h-7" />;
```
oshi97, line 26: can move this out of the body of the component (think I prefer that a little here)

packages/studio/src/components/VersionNumber.tsx
```diff
@@ -0,0 +1,58 @@
...
+    top: "50%",
+    left: "50%",
+    right: "50%",
+    bottom: "auto",
+    marginRight: "-95%",
+    maxWidth: "600px",
+    transform: "translate(-50%, -50%)",
+  },
+  overlay: {
+    zIndex: 1000,
```
oshi97, line 18: do we specify z-index in the other modals?

alextaing: I think we do in all our custom modals

oshi97: oh nevermind I see this is copied from Modal.tsx, could we turn the current Modal.tsx into DialogModal.tsx (or something), and have it use a more general, basic modal component that we also use here??

oshi97: oop didn't see your message in the middel

alextaing: done!

packages/studio/src/components/VersionNumber.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <p className="font-bold mb-4"> Studio Info</p>
+    Studio Verison: {version}
+  </div>
+);
+
+/**
+ * Renders the current version number
+ */
+export default function VersionNumber() {
+  const renderModal: renderModalFunction = useCallback(
```
oshi97, line 19: we can move this outside the component body!

alextaing: oh, so true

oshi97 (approved):
🔥 🔥 🔥

oshi97 (approved):
🌊 🌊 🌊

PR #349, opened 2023-08-24
Ensure that `immer` is installed along with `@yext/studio`.

tmeyer2115:
The `studio` package has a direct dependency on the `immer` 3rd party library. We accidentally never added it to the `package.json` though. This wasn't an issue for anyone using `@yext/studio-plugin` v0.16 or less. In those older versions of `@yext/studio-plugin`, we used the `react-dev-utils` library, which has `immer` as a transitive dependency. So, we got lucky. As of v0.17 of `@yext/studio-plugin`, though, we got rid of `react-dev-utils`.


alextaing:
Studio breaks for any immer version above `immer@9.0.19`.  The current version studio used before adding `immer` to studio's dependencies was `9.0.19`.

nmanu1 (approved):

PR #350, opened 2023-08-25
Set up requests to the Management API

nmanu1:
This PR sets up requests to the Management API. Because the Management API doesn't currently pass the `Access-Control-Allow-Origin` header with the response, we're not able to make requests directly from the browser. Instead, `studio` will need to send a message to `studio-plugin` to fire off those requests using `cross-fetch`.

Also, because we don't know how we'll get the user's API key yet, this functionality is turned off for now (see [this](https://github.com/yext/studio/pull/350/commits/a4bba34709b8ec6de1d6c26773f219214906f5d7) commit for a partial example of how it would be added). Lastly, this PR fixes a bug in `registerListener` so that it correctly awaits the `handler` before returning. This way, a rejected promise is actually caught by the `catch` block.

Because there's discussion around updating the Management API to pass the correct CORS header in the response, we're hopeful that this code can soon be updated so the requests are made directly from the browser. To make that switch easier, I've set up the message handling so there is a message corresponding with each API endpoint method. This way, most of the data aggregation logic will still live in `studio`, rather than `studio-plugin`. This means we're sending some extra messages to `studio-plugin`, but I did some experimenting and the majority of the time to aggregate the data is due to the API requests rather than the message passing within Studio, so there won't be too much of a performance difference (the extra time is ~350ms for the Slapshot Test account). The plan is to have most of this aggregation occur asynchronously on start up of the Studio store, so it shouldn't affect the load time of the stream scope modal.

J=SLAP-2861
TEST=manual

Manually supply an API key and see that I can make requests from Studio to get all the information we need to populate the stream scope modal.


packages/studio-plugin/src/http/HttpService.ts
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * HttpService is a wrapper around the native implementation of AJAX-related
+ * matters.
+ */
+export class HttpService {
+  /**
+   * Performs a GET request.
+   */
+  static async get(
```
oshi97, line 27: Maybe dumb question but how come this works but calling from the browser doesn't? Wouldn't the origin when called from the node process just be like... the user's IP address or something, which would still fail the cross-origin check?

packages/studio-plugin/src/http/ManagementApiService.ts
```diff
@@ -0,0 +1,82 @@
...
+} from "./types/Entities";
+import { BaseQueryParams } from "./types/QueryParams";
+import { SavedFilterData } from "./types/SavedFilters";
+import ResponseParser from "./utils/ResponseParser";
+
+/**
+ * ManagementApiService handles making requests to the Management API for an
+ * account that matches the given API key.
+ */
+export default class ManagementApiService {
```
oshi97, line 16: We don't need to do this now, but would be nice to make this a separate package. Currently if we did that I think user's would need to install yet another dependency which we probably don't want. 
Eventually I think we'll figure out how to allow user's to get studio with just one npm package and then we could move this to a separate package kind of like search-core?

packages/studio-plugin/src/http/constants.ts
```diff
@@ -0,0 +1,10 @@
+export const DEFAULT_API_VERSION = 20220511;
+export const MANAGEMENT_API_ENDPOINTS = {
+  entities: "https://api.yextapis.com/v2/accounts/me/entities",
```
oshi97, line 3: nit we can use a template string for the shared domain https://api.yextapis.com

packages/studio-plugin/src/messaging/registerGetEntityTypes.ts
```diff
@@ -0,0 +1,28 @@
...
+    try {
+      const response = await managementApiService.getEnabledEntityTypes();
+      return {
+        msg: "Successfully fetched entity types.",
+        type: ResponseType.Success,
+        entityTypes: response,
+      };
+    } catch {
+      return {
+        msg: "Error fetching entity types.",
```
oshi97, line 23: Do we want to include the original error in some way?

packages/studio-plugin/src/createStudioPlugin.ts
```diff
@@ -52,6 +53,11 @@ export default async function createStudioPlugin(
   const localDataMappingManager = studioConfig.isPagesJSRepo
     ? new LocalDataMappingManager(studioConfig.paths.localData)
     : undefined;
+  // TODO: Remove false-gate and pass real API key once we get it
+  const managementApiService =
+    false && studioConfig.isPagesJSRepo
+      ? new ManagementApiService("apiKey")
+      : undefined;
```
oshi97, line 60: I think we can remove this block of code for now, it shouldn't be hard for somebody to come back to this PR and get it again
```  
const managementApiService =
    false && studioConfig.isPagesJSRepo
      ? new ManagementApiService("apiKey")
      : undefined;
```

packages/studio-plugin/src/http/HttpService.ts
nmanu1: I think having the correct CORS headers is only enforced by the browser, because even if I do a `curl` with the exact headers copied from the browser request (including the origin), the request is successful. So, the node process is able to get the response because there's no browser to intercept it

packages/studio-plugin/src/http/ManagementApiService.ts
nmanu1: I also like the idea of it being separate. That's one of the reasons why I wanted `http` to have its own `types` and `constants.ts`

packages/studio-plugin/src/messaging/registerGetEntityTypes.ts
nmanu1: I didn't think it was needed because these requests are purely internal. The user doesn't really know that we're making them. At most, I was thinking we'd display a generic error/warning message if they open the stream scope modal and we don't have some information populated from their account because we got some error. I don't think we'd take any specific action based on particular error messages, and the user likely wouldn't be able to either

packages/studio-plugin/src/createStudioPlugin.ts
nmanu1: Sure, I removed it from `createStudioPlugin` for now, but I left in the `configureStudioServer` logic

packages/studio-plugin/src/messaging/registerGetEntityTypes.ts
oshi97: Maybe we would want to console error in the terminal? It doesn't need to be something the user sees unless they look for it

packages/studio-plugin/src/types/messages.ts
```diff
@@ -63,4 +71,19 @@ export type ResponseEventMap = {
   [MessageID.GenerateTestData]: (BaseResponse | ErrorResponse) & {
     mappingJson: Record<string, string[]>;
   };
+  [MessageID.GetEntities]:
+    | ErrorResponse
+    | (BaseResponse & {
```
oshi97, line 76: Is there a better way to represent these types?

nmanu1: The only way I could think of to simplify the types is to allow any errors that are thrown to propagate up to `studio` via `FatalErrorResponse` instead of catching them in the listener handlers and returning an `ErrorResponse`. But, this would make the invoking code we'll add in `studio` more complicated because we'll have to catch the errors there. If you have a strong preference though, I can switch to that

packages/studio-plugin/src/messaging/registerGetEntityTypes.ts
nmanu1: Updated!

packages/studio-plugin/src/types/messages.ts
oshi97: Maybe we can have some kind of generic?

nmanu1: I thought about a generic, but I was having a hard time coming up with a name that made sense. We have two types of responses with data right now, one where the `ErrorResponse` also returns the data (e.g. for `GenerateTestData`) and one where it doesn't (e.g. endpoints from this PR). What do you think of `ErrorOrResponseWithData` for the latter?

oshi97: Maybe `ManagementApiResponse`? It's a little specific but I think we don't have to be super generic here (which I guess is a little ironic?)

oshi97 (approved):

PR #352, opened 2023-08-25
Investigate duplicate watcher

alextaing:
We still had `buildStart` in `createStudioPlugin` which did not seem to be used, since we have HMR implemented using chokidar.  This PR removes `buildStart`.  I checked that edits to components, templates, and site settings were still functional, and that tests were not affected.

J=SLAP-2904
TEST=manual


oshi97 (approved):
🥳

PR #355, opened 2023-08-29
Use studio-plugin version in package.json instead of wildcard

alextaing:
In npm workspaces, you can use `*` to point to a local dependency (i.e. another package), which is how we add `studio-plugin` as a dependency of `studio`.  This dependency is preserved as a symlink in the root `package-lock.json`.  Since the dependency version of `studio-plugin` in the `studio` `package.json` was `*`, there was not a concrete version of `studio-plugin` specified, so users had to explicitly install `studio` and `studio-plugin`.  This PR finds that we can still specify a version of `studio-plugin` in the `studio` `package.json` and still preserve the symlink for local development.

J-SLAP-2895
TEST=manual

### DIFF
--- a/packages/studio/package.json
+++ b/packages/studio/package.json
@@ -20,7 +20,7 @@
     "@minoru/react-dnd-treeview": "^3.4.1",
     "@restart/ui": "^1.5.2",
     "@vitejs/plugin-react": "^4.0.4",
-    "@yext/studio-plugin": "0.19.0",
+    "@yext/studio-plugin": "0.20.0",
     "autoprefixer": "^10.4.14",
     "cac": "^6.7.14",
     "classnames": "^2.3.2",
--- a/packages/studio/package.json
+++ b/packages/studio/package.json
@@ -20,7 +20,7 @@
     "@minoru/react-dnd-treeview": "^3.4.1",
     "@restart/ui": "^1.5.2",
     "@vitejs/plugin-react": "^4.0.4",
-    "@yext/studio-plugin": "*",
+    "@yext/studio-plugin": "0.19.0",
     "autoprefixer": "^10.4.14",
     "cac": "^6.7.14",
     "classnames": "^2.3.2",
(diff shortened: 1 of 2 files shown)


oshi97:
very pog, please make a corresponding item!

packages/studio/package.json
```diff
@@ -20,7 +20,7 @@
     "@minoru/react-dnd-treeview": "^3.4.1",
     "@restart/ui": "^1.5.2",
     "@vitejs/plugin-react": "^4.0.4",
-    "@yext/studio-plugin": "*",
+    "@yext/studio-plugin": "0.19.0",
```
oshi97, line 23: out of curiosity now that the versions are 0.20.0, if we forget to update it here and keep it at 0.19.0 what happens?

packages/studio/package.json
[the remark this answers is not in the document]
alextaing: I was thinking that it might be nice to keep them as `*` so we don't have to update them every release.

packages/studio/package.json
alextaing: The symlink is still preserved in the package-lock, but the test-site will still be using the old version (which in this case caused an error since 0.19.0 is missing the TailwindPropInput stuff). This because the test site only has `studio` as a dependency, which still has the old `studio-plugin` version as a dependency.

packages/studio/package.json
```diff
@@ -20,7 +20,7 @@
     "@minoru/react-dnd-treeview": "^3.4.1",
     "@restart/ui": "^1.5.2",
     "@vitejs/plugin-react": "^4.0.4",
-    "@yext/studio-plugin": "0.19.0",
+    "@yext/studio-plugin": "0.20.0",
```
nmanu1, line 23: do we want to pin to a specific version, or can we add a caret here?

alextaing: I was thinking since we're releasing a lot of versions at the moment and some of them have had breaking changes, it might be a good idea to pin specific versions.

alextaing: Since we're going through a lot of versioning right now that might have a lot of unexpected changes, I was thinking it would be safer to pin versions.  I think this would prevent possible influx of issues from mixing and matching studio and plugin versions.  But I don't have too much experience with releases and versioning-- what are your thoughts?

oshi97 (approved):
lgtm!!

PR #356, opened 2023-08-29
Release v0.20.0 of Studio & Studio Plugin.

tmeyer2115:
## Features
- Studio now supports a user's Tailwind customizations for background color, text color, and font size. Specifically, there is a new Prop Editor that lets them add & remove these custom utility classes. They can add multiple utility classes at once. (#323, #345)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22768,7 +22768,7 @@
     },
     "packages/studio": {
       "name": "@yext/studio",
-      "version": "0.19.0",
+      "version": "0.20.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


oshi97:
I think a directive in tailwind specifically means [things like `@tailwind components`](https://tailwindcss.com/docs/functions-and-directives) that you put in your css file and not the individual css classes
I would probably call them "classes" or "utility classes" or even "class names"? It always makes me a little uneasy using the word classes here because it's such a loaded term, but it's more or less what they call them in the docs https://tailwindcss.com/docs/utility-first

oshi97 (approved):

PR #357, opened 2023-08-29
Parsing Speedup by Adding Source Files via tsconfig

oshi97:
This PR updates our parsing classes to avoid dynamically add source files to the project during initial parsing.
Instead, we specify a path to the user's tsconfig, so that ts-morph can add them all at once on startup.
Originally, I wanted to never dynamically add source files. However, this is still necessary for when we add new files,
at least not without reworking at least a good portion of how we write to file.

This speeds up initial parsing  because every time a source file is added to the project (for instance via `addSourceFileAtPath` like we were doing), ts-morph will recreate the entire typescript compiler Project instance, which is a slow operation.

I update the unit test workflow to split up the studio/studio-plugin tests, and also turn off coverage collection to speed up tests. We can turn it back on when we actually do something with it.
We can also drop down the playwright test timeout now. The performance improvements here help, and also
when we [updated the pages dev startup to use the api](https://github.com/yext/studio/pull/348) instead of a spawnSync.

J=SLAP-2909
TEST=manual

my local test-site has speed up to ~6 second start time, down from ~18 seconds


.github/workflows/run-tests.yml
```diff
@@ -2,21 +2,26 @@ name: Run Tests
 on: workflow_call
 
 jobs:
-  unit_tests:
+  studio:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest]
-        node-version: [18.x]
```
tmeyer2115, line 9: For my own edification, why is the `node-version` no longer needed?

packages/studio-plugin/tests/__utils__/createTestSourceFile.ts
```diff
@@ -1,13 +1,21 @@
-import { createTsMorphProject } from "../../src/ParsingOrchestrator";
+import { Project } from "ts-morph";
 
 export default function createTestSourceFile(
```
tmeyer2115, line 3: Looking at this method again, it's a bit confusing. It does more than simply create a SourceFile, it also creates a new Project. If we were to have this method take in a `ts-morph` project, would that speed up our unit tests at all? Or not really since when you add a SourceFile it recreates the TypeScript Compiler Project anyway. My thought was in a test file, such as `TypeNodeParsingHelper.test.ts`, we could have 
```
const globalProject = createTestProject(); 
```
at the top of the file. Then, in the individual tests have `createTestSourceFile([whatever Code], globalProject, [optional filePath[)`.

.github/workflows/run-tests.yml
oshi97: we default to 18 in setup-ci

packages/studio-plugin/tests/__utils__/createTestSourceFile.ts
oshi97: I think it would speed up unit tests that share the same source files, but unit tests that call `createTestSourceFile` call it to test some individual bit of code, so I doubt there would be performance benefit from sharing the Project instance

tmeyer2115 (approved):

PR #358, opened 2023-08-30
Wait for images to load on Playwright Tests

alextaing:
Images in the Playwright tests were not completely rendered before screenshots were taken.  This really only affected our tests for `add-element` and `rearrange-elements`.  This PR changes all full-page and preview screenshots to check for iFrame images to load before capturing the screenshot.

J=SLAP-2913
TEST=auto, manual


alextaing:
300s timeout appears reliable locally (200 ms had a few failures).  Decided on 1s timeout to be safe.

e2e-tests/tests/infra/StudioPlaywrightPage.ts
```diff
@@ -233,4 +233,26 @@ export default class StudioPlaywrightPage {
   getComponentPath(componentName: string) {
     return path.join(this.tmpDir, "src/components", componentName + ".tsx");
   }
+
+  async waitForIFrameImagesToLoad() {
+    const images = await this.preview.getByRole("img").all();
+    for (const img of images) {
```
oshi97, line 239: It shouldn't really matter but can we use Promise.all here?

e2e-tests/tests/add-element.spec.ts
```diff
@@ -7,14 +7,14 @@ const expectedPage = fs.readFileSync(
...
 );
 
-studioTest("can add a Footer component", async ({ page, studioPage }) => {
+studioTest("can add a Footer component", async ({ studioPage }) => {
   await studioPage.switchPage("LocationPage");
   await studioPage.addElement("Footer", "Components");
   const previews = studioPage.preview.getByText("© 2023 Yext");
   await expect(previews).toHaveCount(1);
-  await expect(page).toHaveScreenshot();
+  await studioPage.takeScreenshotAfterImgRender();
```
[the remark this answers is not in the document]
alextaing: Should we make another item to go through and tighten the scopes on the screenshots, so that they're not all whole page captures?

If we do this, though, I think that we might need to add some more tests to make sure that we keep the same coverage. For example, we might need to add a test for the save button graying out after a save, depending on how we tighten these screenshots.

e2e-tests/tests/infra/StudioPlaywrightPage.ts
```diff
@@ -233,4 +233,26 @@ export default class StudioPlaywrightPage {
...
+          {
+            message: "Wait for images in page preview to render.",
+            timeout: 1000,
+          }
+        )
+        .toBeTruthy();
+    }
+  }
+
+  async takeScreenshotAfterImgRender() {
```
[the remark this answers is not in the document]
alextaing: Ya I went back and forth on this-- but yeah, I think that sounds good.  I replaced all calls to take a full page screenshot with this version where we wait for images to load first.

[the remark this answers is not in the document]
alextaing: I also did the same where we take screenshots of Studio Preview.

e2e-tests/tests/infra/StudioPlaywrightPage.ts
```diff
@@ -233,4 +233,35 @@ export default class StudioPlaywrightPage {
...
+          },
+          {
+            message: "Wait for images in page preview to render.",
+            timeout: 1000,
+          }
+        )
+        .toBeTruthy();
+      imgPromises.push(pollForImageReady);
+    }
+    await Promise.all(imgPromises);
```
oshi97, line 255: what you have def works! but you can use .map to simplify a little
```ts
const imgPromises = images.map(i => {
  expect.poll(() => img.evaluate(blahblahbalb))
})
await Promise.all(imgPromises
```

alextaing: 🧠

e2e-tests/tests/infra/StudioPlaywrightPage.ts
```diff
@@ -233,4 +233,32 @@ export default class StudioPlaywrightPage {
   getComponentPath(componentName: string) {
     return path.join(this.tmpDir, "src/components", componentName + ".tsx");
   }
+
+  async waitForIFrameImagesToLoad() {
+    const images = await this.preview.getByRole("img").all();
+    const imgPromises = images.map((img) =>
+      expect
+        .poll(
+          async () => {
```
oshi97, line 242: I'm pretty sure you don't need the async + await here, it can just be `.poll(() => img.evaluate(...)`

alextaing: I tried that and it works without the async/await, but I wasn't entirely sure about how the mechanics of that worked.  I thought that `expect.poll` keeps polling the function provided, and only passes when (in this case) the function returns `true`.  But `image.evauate` returns a boolean promise, so I wasn't sure how exactly it was working.  Does it already know to wait for the promise to resolve?

oshi97: Oh hmmm it looks like the callback can return a promise?
image
I just assumed it would work because 
`async () => { return await Promise.resolve() }` is equivalent to `async () => { return Promise.resolve }`

oshi97: so if your old code was working this would have to work too

alextaing: ahh interesting, gotcha

e2e-tests/tests/infra/StudioPlaywrightPage.ts
```diff
@@ -238,15 +238,10 @@ export default class StudioPlaywrightPage {
...
-        .poll(
-          () => {
-            return img.evaluate((e) => (e as HTMLImageElement).complete);
-          },
-          {
-            message: "Wait for images in page preview to render.",
-            timeout: 1000,
-          }
-        )
+        .poll(() => img.evaluate((e) => (e as HTMLImageElement).complete), {
```
nmanu1, line 241: instead of casting `e` to HTMLImageElement, does it work if you just set that as the type (i.e. `img.evaluate((e: HTMLImageElement) => e.complete)`)?

e2e-tests/tests/infra/StudioPlaywrightPage.ts
```diff
@@ -233,4 +233,27 @@ export default class StudioPlaywrightPage {
   getComponentPath(componentName: string) {
     return path.join(this.tmpDir, "src/components", componentName + ".tsx");
   }
+
+  async waitForIFrameImagesToLoad() {
```
nmanu1, line 237: nit: can this be private?

oshi97 (dismissed):

PR #359, opened 2023-08-30
Add dropdowns to stream scope modal

nmanu1:
This PR updates the modal for specifying the stream scope in the add page flow so the inputs have a similar UI to the custom Tailwind class input. Each input will have an icon that opens a dropdown with the possible options that can be added for setting the scope. So far, the dropdowns for entity types and saved filters will be correctly populated from the store (tested with dummy data for now). The remaining subtasks will populate the store with actual account content and populate the dropdown for entities, which is more complex.

Also, make the disabled behavior for the Tailwind class prop input more robust: 1) the embed icon is no longer clickable when the input is disabled and 2) the disabled state is more visibly different from the enabled state.

J=SLAP-2905, 2910
TEST=auto, manual

Check that a new page can be created with a stream scope with zero, one, or multiple filters. Makes sure that the disabled state works correctly so that selecting one type of filter (e.g. entity type) will disable the other inputs (saved filters and entities).


packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
```diff
@@ -1,58 +1,185 @@
...
+    return (
+      <>
+        <div className="italic mb-4">
+          Use one of the optional fields below to specify which entities this
+          page can access.
+        </div>
+        {Object.entries(scopeFormData).map(([field, data]) => {
+          const selectedItems: string[] | undefined = streamScope?.[field];
+          const hasOtherScopeFilters =
+            totalStreamScopeItems > (selectedItems?.length ?? 0);
```
tmeyer2115, line 89: To determine if there are other Scope Filters, could we just check if `streamScope` is non-empty?

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
```diff
@@ -1,58 +1,185 @@
...
+          return (
+            <ScopeField
```
tmeyer2115, line 91: Nit: Could we call this `ScopeFilter`, `FilterField`, or something like that? I think it'd be nice if we were closer to the`TemplateConfig` terminology here.

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
```diff
@@ -1,58 +1,185 @@
...
+}
+
+function ScopeField({
+  field,
+  description,
+  allItems,
+  tooltip,
+  emptyText,
+  selectedItems,
+  updateScopeItems,
```
tmeyer2115, line 138: I think we could use a clearer name here. A Stream Scope itself doesn't have items. Maybe `updateFilterValue` or `updateFilterFieldItems`?

packages/studio/src/components/PageSettingsButton/EntityPageModal.tsx
```diff
@@ -9,13 +9,32 @@ import StreamScopeParser, {
...
 import { PageSettingsModalProps } from "./PageSettingsButton";
 import { StaticPageSettings } from "./StaticPageModal";
-import { streamScopeFormData } from "../AddPageButton/StreamScopeCollector";
 import PageDataValidator from "../../utils/PageDataValidator";
 import { toast } from "react-toastify";
 import { isEqual } from "lodash";
 
 type EntityPageSettings = StaticPageSettings & StreamScopeForm;
 
+export const streamScopeFormData: FormData<StreamScopeForm> = {
```
tmeyer2115, line 18: Made an item for the Edit Flow piece: https://yexttest.atlassian.net/browse/SLAP-2918. Could we add a TODO in this file?

packages/studio/src/components/PillPicker/PillPicker.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const toggleOpen = useCallback(() => {
+    if (disabled) {
+      return;
+    }
+    setIsOpen((isOpen) => !isOpen);
+  }, [disabled]);
+
+  const handleSelect = useCallback(
+    (item: string) => {
+      toggleOpen();
```
tmeyer2115, line 38: I think we may want to keep the dropdown open to allow them to select multiple things in one go. This situation is a bit different than the Insert button. In that case, we assumed adding multiple Components in one go was uncommon. For Stream Scope selection, I think adding multiple entity ids or entity types will be common.

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
```diff
@@ -1,58 +1,185 @@
...
-      handleConfirm={onConfirm}
+      handleConfirm={handleConfirm}
+      body={modalBodyContent}
+      confirmButtonText="Next"
+      isConfirmButtonDisabled={false}
     />
   );
 }
+
+interface ScopeFieldProps {
```
tmeyer2115, line 119: I'm getting a bit concerned about Prop bloat here. This Component has a lot of Props. I think that's a bit of a code smell and indicates we may need to refactor. I don't have a great recommendation yet, still thinking.

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
```diff
@@ -1,58 +1,185 @@
...
+  allItems: string[];
+  tooltip: string;
+  emptyText: string;
+  selectedItems: string[] | undefined;
+  updateScopeItems: (selectedItems: string[] | undefined) => void;
+  getDisplayName?: (item: string) => string;
+  disabled?: boolean;
+}
+
+function ScopeField({
```
tmeyer2115, line 131: Could we put this in its own file, maybe in `src/components/common`? We'll want to reuse this when we update the Edit Flow.

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
```diff
@@ -1,58 +1,185 @@
...
+    if (streamScope === undefined) {
+      setStreamScope({});
+    }
+  }, [setStreamScope, streamScope]);
+
+  const updateScopeItems = useCallback(
+    (field: string) => (selectedItems: string[] | undefined) => {
+      const updatedStreamScope = {
+        ...streamScope,
+        [field]: selectedItems,
```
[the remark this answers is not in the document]
nmanu1: I was under the impression that Andrew didn't want a *non*-technical Studio user to be able to specify a "combined" filter for the scope. The disabling behavior in the UI that you mentioned should be sufficient to prevent that use case, particularly for the Add Page flow where the stream scope would never have any pre-existing filters defined.

For editing existing pages' template configs, if we wanted to stop technical users, we'd have to add a check while parsing their template config and have it give an error if filters of multiple types are specified. Technically, this particular function's logic wouldn't stop you from defining a "combined" filter. But, because I wasn't sure how we'd want to handle the technical user's case for existing pages, I thought it would be more graceful to leave in the merging of the existing stream scope.

But, since updating the Page Settings modal will be another item, I can also update this logic to strictly enforce the XOR for now if you prefer

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
nmanu1: No, that would disable all `ScopeField`s once a single filter was added to the stream scope. We only want to check if other scope types have filters (i.e. there are more filters for the scope beyond the ones for this particular "type") so users can continue adding filters of the same type

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
```diff
@@ -1,58 +1,185 @@
...
+      <Tooltip anchorSelect={`#${labelId}`} content={tooltip} />
+      <PillPickerInput
+        selectedItems={selectedItems}
+        availableItems={availableItems}
+        selectItem={addItem}
+        removeItem={removeItem}
+        getDisplayName={getDisplayName}
+        disabled={disabled}
+        customContainerClasses="mt-2 mb-4 border-gray-500"
+        emptyText={emptyText}
```
oshi97, line 171: Would it be possible to write this component in a way where it takes in less props?

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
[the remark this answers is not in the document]
tmeyer2115: Whoops, I meant non-technical user! If you don't mind, let's enforce the XOR here as well. I think it's always good to program defensively if we can. Someone may inadvertently break the disabling logic below, but we'd still be safe if we enforce it here as well. If a technical person has a more complex, "combined" Scope filter, I think we shouldn't allow the non-technical user to edit it the Scope in any way. Whenever a non-technical person is able to make an edit, we should enforce they're only adding one type of filter.

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
tmeyer2115: Gotcha!

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
nmanu1: Changed it to `ScopeFilterField`

packages/studio/src/components/PillPicker/PillPicker.tsx
nmanu1: Updated! This also affects the dropdown for the Tailwind class prop input, but I don't think that's an issue since I could see people wanting to add multiple Tailwind classes at once

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
nmanu1: Changed to `updateFilterFieldItems`

packages/studio/src/components/PillPicker/PillPicker.tsx
```diff
@@ -0,0 +1,96 @@
...
+        onClick={toggleOpen}
+        className={embedIconClasses}
+        aria-label="Toggle pill picker"
+      />
+      {isOpen && renderDropdown(availableItems, selectItem, getDisplayName)}
+    </div>
+  );
+}
+
+function renderDropdown(
```
alextaing, line 54: Maybe not a right-now issue, but would it be useful to have this dropdown and the field picker dropdown make use of a shared dropdown component?

alextaing, line 54: Maybe not a right-now issue, but would it be useful to have this dropdown and the field picker dropdown make use of a shared dropdown component?

nmanu1: Right now, this dropdown is much simpler than the field picker dropdown (which has recursive, nested dropdowns). But, in SLAP-2907, we're going to need a more complex dropdown for entities, so it may make sense to make a shared dropdown as part of that sub-task. I'll call that out in its description!

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
nmanu1: I simplified the prop interface so it's more manageable!

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
nmanu1: I gave it some thought and consolidated the logic to reduce a couple props, but I'm having a hard time reducing it any further. A couple things I considered but didn't end up doing:
- Moving the `empty` display case into the parent components (to remove `emptyText`): This is definitely doable, but the logic is pretty similar for both the Tailwind prop input and stream scope input, so moving it into the parents involved a little code duplication. Also, the `pillContainerClass` styling would need to be exported and used by both parent components, and they'd also have to handle adding the gray "disabled" background, which looked a little messy
- Moving the display name logic fully into `ScopeFilterField` (to remove `getDisplayName`): This is also possible, but we'd have to do some kind of regex matching or deconstructing of the display string to extract the `id` when updating the stream scope, which didn't seem ideal

e2e-tests/src/components/CustomTailwindButton.tsx
```diff
@@ -4,6 +4,8 @@ interface Props {
   className?: TailwindClass;
 }
 
+export const initialProps: Props = { className: "" };
```
oshi97, line 7: Is this necessary?

packages/studio/src/components/PageSettingsButton/EntityPageModal.tsx
```diff
@@ -9,13 +9,33 @@ import StreamScopeParser, {
...
 import { StaticPageSettings } from "./StaticPageModal";
-import { streamScopeFormData } from "../AddPageButton/StreamScopeCollector";
 import PageDataValidator from "../../utils/PageDataValidator";
 import { toast } from "react-toastify";
 import { isEqual } from "lodash";
 
 type EntityPageSettings = StaticPageSettings & StreamScopeForm;
 
+// TODO (SLAP-2918): Update modal to have stream scope dropdowns
+export const streamScopeFormData: FormData<StreamScopeForm> = {
```
oshi97, line 19: It doesn't look like this variable is used in this file so maybe it should live somewhere else?

e2e-tests/src/components/CustomTailwindButton.tsx
nmanu1: I added that so the custom tailwind test doesn't need an extra step to set the `className` prop to a default (instead of undefined/disabled) using the undefined menu button. We'd have to do one or the other, and this one seemed slightly simpler? Not a strong preference though

packages/studio/src/components/PillPicker/PillPickerInput.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { ReactComponent as X } from "../../icons/x.svg";
+import classNames from "classnames";
+import PillPicker from "./PillPicker";
+
+const pillContainerClass =
+  "flex flex-wrap min-h-[38px] items-center border border-gray-400 rounded-lg pt-2 pb-1 px-2 w-full text-sm";
+
+interface PillPickerInputProps {
+  selectedItems: string[] | undefined;
+  availableItems: string[] | undefined;
```
oshi97, line 11: I think it might be simpler to have an items prop like `items: { value: string, displayName: string, isSelected: boolean }[]`? Otherwise it's a little hard to understand how all 3 of these props interact

e2e-tests/src/components/CustomTailwindButton.tsx
oshi97: Oh gotcha we can keep it!

packages/studio/src/components/PageSettingsButton/EntityPageModal.tsx
nmanu1: It's used on line 67. We used to import it before, but now it's more specific to this file, so I moved it here

packages/studio/src/components/AddPageButton/StreamScopeCollector.tsx
oshi97: It looks a lot better!

packages/studio/src/components/PillPicker/PillPickerInput.tsx
nmanu1: I think a single `items` prop would simplify some stuff, but it's also slightly more restrictive than having separate props. For example, if a developer has specified classes in a Tailwind class prop that aren't part of a user's custom Tailwind config, I think the current approach would still show the classes as expected and allow them to be removed, but not added again from the dropdown. By combining the props into a single `items` prop, we'd no longer be able to support developer-specified selected items that come from outside of the list of items that are shown in the dropdown.

We could do an intermediate approach, where `selectedItems` and `availableItems` have type `{ value: string, displayName?: string }`, and still support the above case. But, this adds complexity to `TailwindPropInput`, which doesn't need the concept of display names. We'd have to map from string to object and back, whereas a separate prop for the display name makes it easier to "opt out" of that complexity.

packages/studio/src/components/common/ScopeFilterField.tsx
```diff
@@ -0,0 +1,63 @@
...
+// is used for page settings
+const emptyTextData: { [field in keyof StreamScope]: string } = {
+  entityIds: "No entities found in the account.",
+  entityTypes: "No entity types found in the account.",
+  savedFilterIds: "No saved filters found in the account.",
+};
+
+export interface ScopeFilterFieldProps {
+  field: string;
+  allItems: {
```
oshi97, line 17: Is there a better name than allItems? I was tracing how we pass this down and there's a lot of layers, and hopefully the name could be more helpful when doing this

nmanu1: Renamed to `filterOptions`

oshi97 (approved):
lgtm!

PR #360, opened 2023-08-31
Update Studio and Studio Plugin to v0.21.0.

tmeyer2115:
## Changes
- Studio's startup time is now considerably faster. This is due to an improvement in the logic for parsing Component and Template files. (#357)
- The Studio Plugin is now a direct, explicit dependency of Studio. That means users only have to install `@yext/studio`. They don't need to worry about manually installing `@yext/studio-plugin` as well. (#355)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22768,13 +22768,13 @@
     },
     "packages/studio": {
       "name": "@yext/studio",
-      "version": "0.20.0",
+      "version": "0.21.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


nmanu1 (commented):
Does #355 allow users to install just `studio` instead of both `studio` and `studio-plugin`? If so, we should add that to the notes

tmeyer2115:
> 

I'm talking with Alex now about that. I also need to know if the way we publish the two libraries needs to change.

nmanu1 (approved):

PR #362, opened 2023-08-31
Version Update Script

oshi97:
J=SLAP-2917
TEST=manual

updates studio and studio-plugin versions to next minor


update-minor-versions.mjs
```diff
@@ -0,0 +1,35 @@
...
+
+function bumpStudio(newVersion) {
+  console.log(`... bumping studio to ${newVersion}`);
+  const packageJsonPath = "./packages/studio/package.json";
+  const packageJson = readJson(packageJsonPath);
+  // `npm i @yext/studio-plugin@${newVersion} --save-exact` does not update the package json,
+  // likely because the new version does not exist yet.
+  packageJson.dependencies["@yext/studio-plugin"] = newVersion;
+  fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2));
+  execSync(`npm version ${newVersion} -w=packages/studio`);
```
nmanu1, line 27: should this be `npm version minor -w=packages/studio` like it is for studio-plugin in case the two packages are on different minor versions?

update-minor-versions.mjs
nmanu1, line 1: is this file `.mjs` because we don't have `type: module` in the top-level `package.json`?

update-minor-versions.mjs
oshi97: I was thinking we'd want them to enforce always being in sync but either works!

update-minor-versions.mjs
oshi97: yup we can either have this be .mjs or add type: module

update-minor-versions.mjs
```diff
@@ -0,0 +1,35 @@
...
+  const packageJson = readJson("./packages/studio-plugin/package.json");
+  const newVersion = packageJson?.version;
+  if (!newVersion) {
+    throw new Error("Could not parse studio-plugin version from package.json");
+  }
+  return newVersion;
+}
+
+function bumpStudio(newVersion) {
+  console.log(`... bumping studio to ${newVersion}`);
```
nmanu1, line 20: nit: if we're not enforcing that the versions stay in sync, then I think we should remove `newVersion` from this log because it could be misleading

nmanu1: oh I just meant changing it to `console.log("... bumping studio")` or something like we have for bumping studio-plugin. I think it's still nice to have it log something, even though we don't know what the version will be

nmanu1 (approved):

PR #364, opened 2023-09-01
Separate yext/studio into yext/studio-ui

oshi97:
This PR separates the react and zustand code into yext/studio-ui.
A future PR will replace the typescript compiler with a bundler, which will reduce round trips done by the vite dev server
and also allow us to minify the code.

generateTailwindSafelist had to be moved into studio-plugin, otherwise postcss would complain about things like `window` not existing. This was due to postcss importing all of studio-ui when it tries to import generateTailwindSafelist, and postcss running on node while studio-ui expecting to be run in the browser. Previously, it would directly
import the ts file, and rely on whatever transpilation postcss/tailwind are doing.

J=SLAP-2919
TEST=manual

dev flow still works - I can make an edit in studio-ui with the dev watch running and the test-site will receive updates
it's a full page reload every time now though


packages/studio/tailwind.config.ts
```diff
@@ -50,6 +50,7 @@ export default {
   content: [
     path.resolve(__dirname, "src/**/*.{ts,tsx}"),
     path.resolve(__dirname, "index.html"),
+    path.join(path.dirname(require.resolve("@yext/studio-ui")), "**/*.js"),
```
alextaing, line 53: Is this intended to be `**/*.js`, or ts/tsx?

oshi97: js!

packages/studio-ui/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "@testing-library/user-event": "^14.4.3",
+    "@types/jest": "^29.2.4",
+    "@types/lodash": "^4.14.191",
+    "@types/node": "^18.11.15",
+    "@types/path-browserify": "^1.0.0",
+    "@types/react": "^18.0.26",
+    "@types/react-dom": "^18.0.10",
+    "@types/react-modal": "3.13.1",
+    "@yext/studio-plugin": "*",
+    "copyfiles": "^2.4.1",
```
alextaing, line 65: I see that we might have some new dev dependencies-- should we regenerate the third party notices to include those, and for studio-ui?

packages/studio-ui/package.json
alextaing, line 1: should we generate a Third Party Notices file for studio-ui?

packages/studio-ui/package.json
```diff
@@ -0,0 +1,73 @@
...
+      "import": "./lib/src/index.js",
+      "require": "./lib/src/index.js"
+    }
+  },
+  "files": [
+    "lib"
+  ],
+  "scripts": {
+    "dev": "npm run copy-svg && tsc --watch --preserveWatchOutput -p tsconfig.json",
+    "copy-svg": "npx copyfiles \"src/**/*.svg\" lib",
```
alextaing, line 19: for my _edification_, I saw that before we didn't include icons in lib-- why do we need them now/didn't we need them before?

packages/studio-ui/src/store/models/StudioStore.ts
```diff
@@ -22,4 +23,5 @@ export type StudioStore = {
   actions: StudioActions;
   studioConfig: StudioConfigSlice;
   studioEnvData: StudioEnvDataSlice;
+  gitData?: GitData;
```
alextaing, line 26: is this something we still want to add right now, even if we don't do anything with it or have a slice for it yet?

packages/studio-ui/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
```
alextaing, line 1: I saw that in the `tsconfig` of studio-plugin, we extend the config from the root `tsconfig` since a lot of it is the same. Is that something that we would want to do for all the `tsconfig`s?  I'm also not entirely clear how this works outside of `studio-prototype` in npm world.

packages/studio-ui/tsconfig.json
alextaing, line 1: I saw that in the `tsconfig` of studio-plugin, we extend the config from the root `tsconfig` since a lot of it is the same. Is that something that we would want to do for all the `tsconfig`s?  I'm also not entirely clear how this works outside of `studio-prototype` in npm world since the configs are extended using relative filepaths.

packages/studio-ui/src/store/models/StudioStore.ts
oshi97: Oh oop sorry I'll remove this -

packages/studio-ui/package.json
oshi97: Oh hmm would you be ok waiting until we update the third party notices script?

packages/studio-ui/package.json
oshi97: oh so typescript doesn't handle .svg imports at all, but we still need these svgs when running the dev build

https://github.com/microsoft/TypeScript/issues/30835

packages/studio-ui/tsconfig.json
oshi97: Oh we extend tsconfigs just to simplify things a little and keep options in sync
not sure why these aren't extending so I can update them!

alextaing:
lgtm 🔥 but @nmanu1, do you want to take a gander?

packages/studio/package.json
```diff
@@ -10,54 +10,23 @@
...
-    "typecheck-jest": "npx tsc -p tests/tsconfig.json"
+    "build:windows": "rimraf lib && tsc -p tsconfig.node.json && tsc -p tsconfig.json"
   },
   "dependencies": {
-    "@dhmk/zustand-lens": "^2.0.5",
-    "@minoru/react-dnd-treeview": "^3.4.1",
-    "@restart/ui": "^1.5.2",
     "@vitejs/plugin-react": "^4.0.4",
     "@yext/studio-plugin": "0.22.0",
+    "@yext/studio-ui": "0.22.0",
```
nmanu1, line 18: to check my understanding, we'll now be publishing `studio-plugin`, `studio-ui`, and `studio` to npm, but user's should only have to install `studio` to run Studio with their repo?

packages/studio/package.json
```diff
@@ -10,54 +10,23 @@
...
     "postcss": "^8.4.27",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "react-modal": "3.16.1",
-    "react-toastify": "^9.1.1",
-    "react-tooltip": "^5.18.0",
-    "tailwind-merge": "^1.8.1",
-    "tailwindcss": "^3.3.3",
-    "vite": "^4.4.7",
-    "vite-plugin-svgr": "^2.4.0",
```
nmanu1, line 40: do we still need `vite` and `vite-plugin-svgr` deps in `studio` for the Vite config?

packages/studio-ui/tsconfig.json
```diff
@@ -0,0 +1,8 @@
...
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "lib",
+    "types": ["@yext/studio-plugin/virtual-module", "vite-plugin-svgr/client"]
```
nmanu1, line 5: does `studio-ui` need this type?

packages/studio-ui/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "@types/react-dom": "^18.0.10",
+    "@types/react-modal": "3.13.1",
+    "@yext/studio-plugin": "*",
+    "copyfiles": "^2.4.1",
+    "jest": "^29.5.0",
+    "jest-environment-jsdom": "^29.3.1",
+    "resize-observer-polyfill": "^1.5.1"
+  },
+  "peerDependencies": {
+    "@yext/studio-plugin": "*"
```
nmanu1, line 71: should this be pinned to a specific version?

packages/studio-ui/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "postcss": "^8.4.27",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-modal": "3.16.1",
+    "react-toastify": "^9.1.1",
+    "react-tooltip": "^5.18.0",
+    "tailwind-merge": "^1.8.1",
+    "tailwindcss": "^3.3.3",
+    "vite": "^4.4.7",
+    "vite-plugin-svgr": "^2.4.0",
```
nmanu1, line 45: `studio-ui` doesn't use Vite, does it?

packages/studio/package.json
```diff
@@ -10,54 +10,23 @@
...
     "zundo": "2.0.0-beta.12",
     "zustand": "^4.3.2"
```
nmanu1, line 26: `studio` doesn't need `zundo` or `zustand` anymore, right?

packages/studio-ui/src/App.tsx
```diff
@@ -3,6 +3,7 @@ import ActionsBar from "./components/ActionsBar";
 import Toast from "./components/Toast";
 import PreviewWithUseComponents from "./components/PreviewWithUseComponents";
 import LeftSidebar from "./components/LeftSidebar";
+import "react-tooltip/dist/react-tooltip.css";
```
nmanu1, line 6: based on the [docs](https://react-tooltip.com/docs/getting-started#set-the-default-styling), I think we shouldn't need to manually import this CSS

packages/studio-ui/tsconfig.json
oshi97: yes

packages/studio/package.json
oshi97: 👍

packages/studio-ui/tsconfig.json
oshi97: it still needs it!

packages/studio/package.json
nmanu1: can we update the `update-minor-versions` script to bump `studio-ui` too?

packages/studio-ui/package.json
oshi97: I don't think we need to

packages/studio/package.json
oshi97: it will need to be added back in the following pr

packages/studio-ui/package.json
oshi97: it will need to be added back in the following pr

update-minor-versions.mjs
```diff
@@ -1,35 +1,47 @@
...
+
+function readJson(filepath) {
+  return JSON.parse(fs.readFileSync(filepath, "utf-8"));
+}
+
+function readVersion(packageJsonPath) {
+  const packageJson = readJson(packageJsonPath);
+  const newVersion = packageJson?.version;
+  if (!newVersion) {
+    throw new Error("Could not parse studio-plugin version from package.json");
```
nmanu1, line 36: nit: can the error message be updated so it's not hard-coded for `studio-plugin`?

oshi97: updated!

alextaing:
Things look good to me! 🥳

PR #365, opened 2023-09-06
Add entity type dropdown to entityIds StreamScope field

oshi97:
This PR adds an entity type dropdown to the entityIds StreamScope field.
It also adds dynamic fetching of entities once the user scrolls down the bottom of the select.

J=SLAP-2907
TEST=manual

I can specify an entity type, select entity ids, then switch entity type and select more ids
saving the page correctly adds the entity ids to the TemplateConfig


packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,125 @@
...
+      updateSelection(selectedItems.map((item) => item.value));
+    },
+    [updateSelection]
+  );
+
+  const onMenuScrollToBottom = useCallback(async () => {
+    if (!entityType) {
+      return;
+    }
+    const currentNumEntities = Object.keys(entitiesRecord[entityType]).length;
```
nmanu1, line 56: should this be `entitiesRecord[entityType].entities.length`?

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,125 @@
...
+    if (currentNumEntities >= entitiesRecord[entityType].totalCount) {
+      return;
+    }
+    setIsLoading(true);
+    await fetchEntities(entityType, Math.floor(currentNumEntities / 50));
```
nmanu1, line 61: why do we use `Math.floor` here? I thought we'd either not need it at all, or need `Math.ceil`?

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,125 @@
...
+    setIsLoading(false);
+  }, [entitiesRecord, entityType, fetchEntities]);
+
+  const className = classNames("flex flex-col mb-1");
```
nmanu1, line 65: nit: can this variable be a little more descriptive of what element it's styling?

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  if (availableEntityTypes.length === 0) {
+    return (
+      <div className={className}>
+        <StreamScopeFieldLabel streamScopeField="entityIds" />
+        <div className="border border-gray-400 min-h-[38px] rounded-lg pt-2 pb-1 px-2 w-full text-sm bg-gray-50 text-gray-500">
```
nmanu1, line 71: is this the same as the disabled pill picker input styling? if so, is there a way to export/reuse it from there?

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,125 @@
...
+          onChange={onEntityTypeChange}
+          disabled={disabled}
+        >
+          {!disabled &&
+            availableEntityTypes.map((entityType) => (
+              <option key={entityType}>{entityType}</option>
+            ))}
+        </select>
+      </label>
+      {!disabled && (
```
nmanu1, line 94: Can we separate out the disabled case's rendering? The conditionals are making the JSX a little difficult to read

packages/studio/src/store/slices/accountContent/createAccountContentSlice.ts
```diff
@@ -19,6 +24,25 @@ const createAccountContentSlice: SliceCreator<AccountContentSlice> = (set) => ({
       console.error("Unable to fetch account content.", e);
     }
   },
+  async fetchEntities(entityType: string, pageNum: number): Promise<void> {
```
nmanu1, line 27: can we add tests for this?

nmanu1 (commented):
Screenshot 2023-09-07 at 3 29 44 PM

Looks like the tooltip positioning needs to be fixed. Also, is there a way to make the styling more consistent between the different inputs?

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,125 @@
...
+        <Select
```
nmanu1, line 95: Could we just use `react-select` instead of `PillPickerInput` for `StreamScopeField` and `TailwindPropInput`? And maybe for our other `select` elements too (e.g. the entity types dropdown in this component and `UnionPropInput`) so they're consistent?

nmanu1, line 95: Is there a way to turn off the search ability for this component? I don't think it makes sense to have the dropdown be searchable when we may not have loaded in all the entities yet

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,125 @@
...
+      </div>
+    );
+  }
+
+  return (
+    <div className={className}>
+      <StreamScopeFieldLabel streamScopeField="entityIds" />
+      <label className="mt-2 text-sm flex flex-col">
+        {!disabled && "Choosing from entities of type:"}
+        <select
```
nmanu1, line 83: I slightly prefer having only a single input for each stream scope field. It seems a little simpler and more consistent that way because the additional complexity would just be in the dropdown. But, if we want two separate dropdowns, maybe the could be next to each other in the same row? So the entity types dropdown would be shorter and appear in the left part of the row, and a longer entities dropdown next to it on the right?

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,125 @@
...
+      </label>
+      {!disabled && (
+        <Select
+          className="my-2"
+          isMulti={true}
+          onChange={onEntityIdChange}
+          value={selectedOptions}
+          options={entityOptions}
+          onMenuScrollToBottom={onMenuScrollToBottom}
+          isLoading={isLoading}
```
nmanu1, line 102: Is it possible for the loading indicator to be at the bottom of the dropdown instead of in the search input? I think that would make it more clear that the user hasn't reached the end of the dropdown and more items are being loaded in

packages/studio/package.json
```diff
@@ -32,6 +32,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-modal": "3.16.1",
+    "react-select": "^5.7.4",
```
alextaing, line 35: should we regenerate third party notices for this new dependency?

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: We could but I think thats out of scope here

packages/studio/src/store/slices/accountContent/utils.ts
```diff
@@ -39,18 +42,20 @@ export async function fetchEntitiesRecord() {
   return typeToEntitiesMap;
 }
 
-async function fetchInitialEntities(entityType: string) {
-  const entitiesResponse = await sendMessage(
+export async function fetchEntities(
```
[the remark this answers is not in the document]
oshi97: I didn't want to have the function be a nooo when we already hit the total count, it seemed a little strange though I can update if you prefer

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,125 @@
...
+      return Object.values(entityData).length > 0;
+    })
+    .map((entry) => entry[0]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [entityType, setEntityType] = useState<string | undefined>(
+    availableEntityTypes[0]
+  );
+  const entityOptions = useEntityOptions(entityType);
+  const selectedOptions = useMemo(
+    () => selectedIds?.map((id) => ({ value: id, label: id })),
```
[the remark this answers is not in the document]
oshi97: I also think it's a little more transparent this way because this is the information you end up writing in your TemplateConfig. If product asks for it I'd make it a separate item

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: I didnt want to pass a decimal, though maybe that works idk

if we have 10 entities loaded we want to ask for page 0 not page 1

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: I think we'd have to override things in react-select which I'd rather make another item. But I'd also like to be more incremental and wait for what product says here

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: How would we only have a single input here? I think that could be nice

packages/studio/src/store/slices/accountContent/utils.ts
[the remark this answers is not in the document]
oshi97: Honestly I'm not sure which I prefer so I opted for the action to be a little simpler 🤷

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,125 @@
...
+export default function EntityIdField({
+  disabled,
+  updateSelection,
+  selectedIds,
+}: Props) {
+  const [entitiesRecord, fetchEntities] = useStudioStore((store) => [
+    store.accountContent.entitiesRecord,
+    store.accountContent.fetchEntities,
+  ]);
+  const availableEntityTypes = Object.entries(entitiesRecord)
```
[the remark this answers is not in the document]
oshi97: At one point I was getting blank options for certain entity types so I added the filter, but I think that was a different issue!

packages/studio/src/store/slices/accountContent/createAccountContentSlice.ts
```diff
@@ -19,6 +24,25 @@ const createAccountContentSlice: SliceCreator<AccountContentSlice> = (set) => ({
...
+    const res = await fetchEntities(entityType, pageNum);
+    if (!res) {
+      return;
+    }
+
+    const entityIdToEntityData: Record<string, EntityData> =
```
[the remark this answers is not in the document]
oshi97: I was a little worried about having extra logic inside a set() and mutating more than necessary with immer, though there's probably nothing to worry about. I can update if you like!

packages/studio/src/store/slices/accountContent/createAccountContentSlice.ts
oshi97: It's not really doing anything except passing through params and we already have tests for fetchInitialEntities, but I can add a test if you'd like

packages/studio/package.json
oshi97: Oh hmm would you be okay waiting until we update the third party notices script?

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: there's less disabled specific logic now

nmanu1 (commented):
Screenshot 2023-09-08 at 1 01 49 PM

- can the text for the entity type dropdown be `text-sm` to match the other inputs?
- looks like the heights of the dropdowns don't match exactly?
- can the outer border edges be rounded more to match the saved filters and entity type dropdowns?

packages/studio/src/store/slices/accountContent/createAccountContentSlice.ts
```diff
@@ -19,6 +24,25 @@ const createAccountContentSlice: SliceCreator<AccountContentSlice> = (set) => ({
...
+      res.entities.reduce((prev, curr) => {
+        prev[curr.id] = curr;
+        return prev;
+      }, {});
+
+    set((state) => {
+      state.entitiesRecord[entityType].entities = {
+        ...state.entitiesRecord[entityType].entities,
+        ...entityIdToEntityData,
+      };
```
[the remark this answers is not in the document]
oshi97: Oh oops it's weird that typescript didn't yell at me here
everything was still working but only by coincidence that Object.values on an array works

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { pillContainerClass } from "../PillPicker/PillPickerInput";
+import classNames from "classnames";
+
+interface Props {
+  disabled: boolean;
+  updateSelection: (selectedIds: string[]) => void;
+  selectedIds?: string[];
+}
+
+export default function EntityIdField({
```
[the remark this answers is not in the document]
oshi97: I can add a test for the scrolling! In general I didn't want to add too many tests to this section of the code since I feel like a lot of it could change in the near future

packages/studio/src/store/slices/accountContent/createAccountContentSlice.ts
nmanu1: I think it would be helpful to make sure the `entitiesRecord` is being updated correctly for new batches of entities

packages/studio/src/components/AddPageButton/EntityIdField.tsx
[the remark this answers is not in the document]
nmanu1: I think I'd want this to be consistent with the saved filters, so we'd either remove the display name from selected saved filters to match, or add the display name here. I agree that having IDs appear in the input make sense because that's what they're really choosing for the template config, but it also makes it harder to compare against the dropdown when adding multiple filters

packages/studio/src/components/AddPageButton/EntityIdField.tsx
nmanu1: I thought the check above would already handle the cases where the number isn't a multiple of 50. Because if less than 50 entities were fetched, that means we should have reached the end and fetched the total number

packages/studio/src/components/AddPageButton/EntityIdField.tsx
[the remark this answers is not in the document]
oshi97: I tried a lot of things including overriding the MenuList component to give it an aria-label so I could select it with the testing library, but it looks like react-testing-library does not support scrolling tests due to jsdom not supporting layout

https://github.com/testing-library/react-testing-library/issues/671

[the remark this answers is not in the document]
oshi97: We can update our acceptance tests later but I don't plan to do that here

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: It's the same thing for more entities right? If we have 50 entities loaded we'd want to ask for page 1 not page 2

oshi97: I think it's a little safer to floor it even if we have another check

nmanu1: Ah I didn't realize both floor and ceil returned the same for whole numbers. It shouldn't make a difference then. Since we're appending to the entities array, either way is a little risky if we didn't have the first check. With floor, we'd risk duplicating entities if a request returned less than 50, and with ceil, we'd risk skipping entities. But again, it shouldn't matter here

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -116,6 +114,12 @@ const selectStyles = {
     borderTopLeftRadius: "0",
     borderBottomLeftRadius: "0",
     borderColor: "rgb(156, 163, 175)",
+    borderTopRightRadius: "0.5rem",
+    borderBottomRightRadius: "0.5rem",
+  }),
+  menu: (base) => ({
+    ...base,
+    fontSize: "14px",
```
nmanu1, line 122: nit: can we also add this `fontSize` to the placeholder text?

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -85,7 +83,7 @@ export default function EntityIdField({
       </div>
       <div className="flex items-center mt-2">
         <select
-          className="disabled:bg-gray-100 border border-r-0 border-gray-400 text-gray-900 focus:ring-blue-500 focus:border-blue-500 py-2 pl-1 pr-1"
+          className="disabled:bg-gray-100 border border-r-0 border-gray-400 text-gray-900 focus:ring-blue-500 focus:border-blue-500 py-2 pl-1 pr-1 text-sm h-[39px] rounded-l-lg"
```
nmanu1, line 86: nit: I think the height should be 38px, not 39px?

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,140 @@
...
+    return (
+      <div className={containerClass}>
+        <StreamScopeFieldLabel streamScopeField="entityIds" />
+        <div
+          className={classNames(
+            pillContainerClass,
+            "mt-2 bg-gray-50 text-gray-500 pb-2"
+          )}
+        >
+          {!disabled && "No entity types found in the account."}
```
nmanu1, line 73: nit: this should be "No entities found in the account."

packages/studio-plugin/src/http/types/Entities.ts
```diff
@@ -12,7 +12,7 @@ export interface EntitiesQueryParams extends BaseQueryParams {
   limit: number;
 }
 
-interface EntityData {
+export interface EntityData {
```
nmanu1, line 15: I don't think we need to export this type anymore

packages/studio/src/components/AddPageButton/EntityIdField.tsx
```diff
@@ -0,0 +1,140 @@
...
+    await fetchEntities(entityType, Math.floor(currentNumEntities / 50));
+    setIsLoading(false);
+  }, [entitiesRecord, entityType, fetchEntities]);
+
+  const containerClass = "flex flex-col mb-4";
+
+  if (availableEntityTypes.length === 0) {
+    return (
+      <div className={containerClass}>
+        <StreamScopeFieldLabel streamScopeField="entityIds" />
```
nmanu1, line 66: I think this label also needs to be in a `div` to have the correct tooltip positioning

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: I'm not sure how we should deal with this kind of thing long term. Maybe we can change entitiesRecord to store a map of entityId to EntityData instead of EntityData[] but that's still kind of dodging issues

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: For some reason 39px seemed to fit better, I think due to the border width?

packages/studio/src/store/slices/accountContent/createAccountContentSlice.ts
oshi97: Yup there's a test now!

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: Oh hmm this message appears when availableEntityTypes has a length of 0, but I guess that no entity types also means no entities at all? Anyways I'll update the message

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: At 100% zoom I prefer 39px but when I zoom in 38px lines up better
I'll just use 38 for now

packages/studio/src/components/AddPageButton/EntityIdField.tsx
oshi97: I passed text-sm to the entire Select

packages/studio/src/components/AddPageButton/EntityIdField.tsx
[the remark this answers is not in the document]
oshi97: Would you be okay deferring this until we get to the corresponding edit page settings updates? I think we might have a better idea at that point how to proceed

packages/studio/src/components/AddPageButton/EntityIdField.tsx
nmanu1: Not sure I follow. When might this be an issue? Do you mean in the case where we forget the first check?

e2e-tests/__screenshots__/darwin/add-entity-page.spec.ts/can-add-an-entity-page-2.png
nmanu1, line 1: I think these screenshots have to be regenerated after the text change

oshi97: oof

oshi97: ⏲️

oshi97: 🍗

nmanu1 (approved):

PR #366, opened 2023-09-06
Populate initial account content

nmanu1:
This PR populates the store with the account content for saved filters, entity types, and the first 50 entities of each type. The `PillPicker` dropdown styling is also updated to be a fixed max height with scrollable content, instead of growing the dropdown to display all options.

J=SLAP-2906
TEST=auto, manual

Hard-code an API key in studio-plugin and see the stream scope modal for adding an entity page in the test site is populated with the entity types and saved filters from the Slapshot test account. Entity types with no entities are not included in the entity types dropdown.


packages/studio/src/store/useStudioStore.ts
```diff
@@ -56,6 +56,10 @@ const useStudioStore = create<StudioStore>()(
 );
 
 void useStudioStore.getState().actions.refreshActivePageEntities();
+void (
+  useStudioStore.getState().studioConfig.isPagesJSRepo &&
+  useStudioStore.getState().accountContent.refreshBaseAccountContent()
+);
```
nmanu1, line 62: Until we add the logic for sourcing the user's API key in `studio-plugin`, we will see the following error in the browser console. If we'd prefer not to have the error appear in the interim, I can remove this call to `refreshBaseAccountContent`

Screenshot 2023-09-06 at 3 38 43 PM

packages/studio/src/store/slices/accountContent/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export async function fetchEntitiesRecord() {
+  const entityTypesResponse = await sendMessage(
+    MessageID.GetEntityTypes,
+    null,
+    { hideSuccessToast: true }
+  );
+  if (entityTypesResponse.type !== ResponseType.Success) {
+    return {};
```
oshi97, line 22: Should we log an error?

packages/studio/src/store/useStudioStore.ts
```diff
@@ -56,6 +56,10 @@ const useStudioStore = create<StudioStore>()(
 );
 
 void useStudioStore.getState().actions.refreshActivePageEntities();
+void (
+  useStudioStore.getState().studioConfig.isPagesJSRepo &&
```
oshi97: Should we guard refreshActivePageEntities with isPagesJSRepo?

packages/studio/src/store/slices/accountContent/createAccountContentSlice.ts
```diff
@@ -0,0 +1,18 @@
+import { SliceCreator } from "../../models/utils";
+import AccountContentSlice from "../../models/slices/AccountContentSlice";
+import { fetchEntitiesRecord, fetchSavedFilters } from "./utils";
+
+const createAccountContentSlice: SliceCreator<AccountContentSlice> = (set) => ({
+  savedFilters: [],
+  entitiesRecord: {},
+  refreshBaseAccountContent: async () => {
+    const savedFilters = await fetchSavedFilters();
```
oshi97, line 9: use Promise.all it actually matters here to not block requests

packages/studio/src/store/slices/accountContent/utils.ts
nmanu1: An error will already be logged in the terminal if there was an error getting the response (e.g. https://github.com/yext/studio/blob/feature/add-page-improvements/packages/studio-plugin/src/messaging/registerGetSavedFilters.ts#L22)

packages/studio/src/store/useStudioStore.ts
nmanu1: It's not necessary because `refreshActivePageEntities` looks for the `pagesJS.entityFiles` part of the active page's state. And since that doesn't exist for pages in a non-PagesJS repo, `activePageEntities` is set to undefined, which is the correct behavior

packages/studio/src/store/slices/accountContent/utils.ts
oshi97: I still think it'd be nice to log an error in the browser. I'm not sure if studio in storm will show the terminal output. When I implemented a similar fetchEntitiesRecord for my item the console.error was also nice when running jest tests to tell me something was wrong

packages/studio/src/store/useStudioStore.ts
oshi97: is there a way we could catch the error?

packages/studio/src/store/slices/accountContent/utils.ts
nmanu1: Added!

packages/studio/src/store/useStudioStore.ts
nmanu1: Updated to catch the error, so now the console has these logs instead:
Screenshot 2023-09-07 at 9 10 03 AM

oshi97 (approved):

PR #367, opened 2023-09-06
Introduce `Layouts` to Studio

alextaing:
This PR introduces Layouts to Studio. Studio now recognizes layouts that appear in `src/layouts` (or a custom dir) and stores their filepaths in a new `layouts` slice.  Layout names are derived from the layout's filename. Tests were updated to reflect changes in studio config. A test was written to check that layout files are found from the `layouts` dir.

The current `LayoutSlice` just contains a `string[]` of the names of the layouts in the layouts dir.  This is simply a place holder, since no layout parsing was implemented yet.  Layout parsing and the LayoutSlice will be fleshed out in future PRs in the epic.

J=SLAP-2923
TEST=auto,manual


packages/studio-plugin/src/ParsingOrchestrator.ts
```diff
@@ -196,6 +204,17 @@ export default class ParsingOrchestrator {
     return this.filepathToFileMetadata;
   }
 
+  private initLayoutNames(): string[] {
```
oshi97, line 207: Is there a separate item for parsing layouts?

alextaing: I believe Tom is making one!

oshi97 (dismissed):

packages/studio-plugin/src/ParsingOrchestrator.ts
```diff
@@ -131,6 +133,11 @@ export default class ParsingOrchestrator {
       if (fileExists) {
         this.pageNameToPageFile[pageName] = this.getPageFile(pageName);
       }
+    } else if (filepath.startsWith(this.paths.layouts)) {
```
nmanu1, line 136: do we need to consider removing deleted layouts from the `layoutNames` here?

packages/studio/src/store/hotReloadStore.ts
```diff
@@ -16,6 +16,8 @@ export default async function hotReloadStore(payload: StudioHMRPayload) {
     case "modules":
       await syncFileMetadata(studioData, payload.file);
       break;
+    case "layouts":
```
nmanu1, line 19: should we be updating the `layouts` in `LayoutSlice` here?

packages/studio-plugin/src/ParsingOrchestrator.ts
alextaing: I thought about that-- I don't think users will have the ability to delete layouts.  Is that the situation that you're talking about, or is there another reason why we would delete layouts from `layoutNames`?

Also, I think that array just a placeholder for now, and we'll do more in depth parsing in later items.

alextaing: I was thinking about that-- I don't think users will have the ability to delete layouts.  Is that the situation that you're talking about, or is there another reason why we would delete layouts from `layoutNames`?

Also, I think that array just a placeholder for now, and we'll do more in depth parsing in later items.

packages/studio/src/store/hotReloadStore.ts
alextaing: yep!  TODO added

packages/studio-plugin/src/ParsingOrchestrator.ts
nmanu1: I was thinking about HMR, where if someone (presumably a developer) deletes a layout file in their repo while running Studio, then we'd want to delete the layout from `layoutNames`, similar to how we do for components (which also can't be deleted through the UI but could be by a developer).

I figured since it should be pretty quick to add, it would be better to add it now, even though the array is just a placeholder, so we don't forget it once we add the full parsing. Especially, since we already have some logic here (unlike for `hotReloadStore`), I could see us changing the existing logic in a future PR and forgetting to consider the deletion case

packages/studio/__mocks__/virtual_yext-studio.ts
```diff
@@ -5,13 +5,15 @@ const mockFilepath = path.join(__dirname, "../tests/__mocks__");
 const mockStudioData: StudioData = {
   pageNameToPageState: {},
   pageNameToErrorPageState: {},
+  layoutNameToLayoutPath: {},
```
nmanu1, line 8: nit: should this be `layouts` for now?

alextaing: oh! yes, I missed that-- was from an older implementation.

nmanu1 (approved):

PR #368, opened 2023-09-08
studio-ui bundling

oshi97:
Bundles studio-ui to reduce multiple round trips made by the browser to the vite dev server.
The browser still makes network requests for things like localData and component files.
One way to reduce perceived lag here is server side rendering studio.
Another is the loading indicator we have planned.

DOMContentLoaded is now 412ms instead of 5.09s. 
There should be a bigger improvement in CBD due to network requests taking longer there compared to local.
5.0MB transferred instead of 14.6MB (could be smaller probably but still yay!)

Removes ts-morph/typescript dependencies from TypeGuards and by extension the ui bundle (they are big packages).
Updates lodash imports to use smaller scoped imports.
Adds vite plugins for bundle size, css in js, dts files, and svg imports.

J=SLAP-2921
TEST=manual

I can start up studio and see less requests in the network tab

old network tab

https://github.com/yext/studio/assets/23005393/022d6199-5fa6-47db-b84b-5e543c729ae3

new network tab
image


packages/studio/src/main.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import ReactDOM from "react-dom/client";
-import { App, hotReloadStore } from "@yext/studio-ui";
-import { StudioHMRPayload, StudioHMRUpdateID } from "@yext/studio-plugin";
```
tmeyer2115, line 4: Out of curiosity, what does importing `StudioHMRUpdateID` from `studio-ui` do in terms of performance?

oshi97: it's a type import so nothing! I tested both ways and specifying it as a type import with the `type` keyword doesn't affect whether or not network requests are made!

If it we were importing say Typeguards instead (i.e. just any non-type) vite would make network requests for every file under `packages/studio-plugin/src/index.ts`. 
This morning (because work is unfortunately somewhat ingrained in my mind) I was coincidentally reading up on vite prebundling, and this behavior can be explained by how vite treats linked packages in a monorepo as source code by default https://vitejs.dev/guide/dep-pre-bundling.html
Pre bundling can be turned on by adding `optimizeDeps.include` for the specific packages you want, which would address this behavior as well. 

I **think** we still want this off though because bundling studio-ui as a separate build step is better than having vite prebundle take time prebundling things before the dev server launches. Pre bundling still takes some amount of time (I haven't tested but I assume it would take the same amount of time as the current build for studio-ui takes which for me is about 10 seconds) and we also have less visibility about the bundle size when we rely on prebundling.

We could turn on prebundling on TOP of bundling ourselves, as a kind of safety net. But it's completely unnecessary and I'm worried would only mask performance issues

oshi97: I misread your comment, I thought you were asking about StudioHMRPayload not StudioHMRUpdateID.
The behavior for StudioHMRUpdateID is the same as described above for any non-type import

PR #373, opened 2023-09-15
Add HMR for layouts and remove top-level fragments

nmanu1:
This PR removes top-level fragments from the component tree of each layout in the layout slice so they won't be displayed in the UI when adding a layout to a new page. HMR is also updated to re-sync layouts in the store when a layout file is changed or a full sync is performed.

J=SLAP-2937
TEST=manual

In the test-site, saw that if `LocationLayout` had a top-level fragment, it was not present in the component tree in the `layouts` record. Also, checked that updating `LocationLayout` while running the test-site would correctly update it's component tree in the store.

### DIFF
--- a/packages/studio/src/store/hotReloadStore.ts
+++ b/packages/studio/src/store/hotReloadStore.ts
@@ -16,7 +16,7 @@ export default async function hotReloadStore(payload: StudioHMRPayload) {
       await syncFileMetadata(studioData, payload.file);
       break;
     case "layouts":
-      // TODO SLAP-2930
+      syncLayouts(studioData);
       break;
     case "pages":
       syncPages(studioData);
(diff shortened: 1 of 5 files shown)


packages/studio/src/store/hotReloadStore.ts
```diff
@@ -81,6 +82,15 @@ function syncPages(studioData: StudioData) {
   });
 }
 
+function syncLayouts(studioData: StudioData) {
+  const layoutNameToLayoutState = removeTopLevelFragments(
```
tmeyer2115, line 86: We added back in the generic param to `removeTopLevelFragments`. We figured we might need it for Layouts. I think we want to change the signature of `removeTopLevelFragments` from `removeTopLevelFragments<T extends PageState>` to `removeTopLevelFragments<T extends PageState | LayoutState>`.

nmanu1: Updated!

tmeyer2115 (approved):

oshi97 (approved):

PR #376, opened 2023-09-20
Layouts in add page flow

alextaing:
This PR adds layout selection to the end of the add page flow.  An e2e test was written to test creating a static page using a simple layout.  Unit tests were updated to test layouts as a part of the add page flow.

J=SLAP-2925
TEST=auto,manual


packages/studio-ui/src/components/AddPageButton/AddPageButton.tsx
```diff
@@ -38,7 +38,16 @@ function AddPageButtonInternal() {
           setStep(FlowStep.GetBasicPageData);
           break;
         case FlowStep.GetBasicPageData:
-          await createPage(pageName, getPathVal, state.streamScope);
+          setStep(FlowStep.SelectLayout);
+          break;
+        case FlowStep.SelectLayout:
+          state.pageName &&
+            (await createPage(
+              state.pageName,
```
tmeyer2115, line 46: Could we do `state.pageName || ""` here and get rid of the `&&` conditional above? I think that reads a bit cleaner. Really, we know the `state.pageName` is set here. You can't get to this part of the flow without it. Ultimately, we should figure out how we can get rid of the empty string defaulting here. But, that doesn't need to be part of this PR. We were doing that defaulting before.

packages/studio-ui/src/components/AddPageButton/AddPageContext.ts
```diff
@@ -1,14 +1,18 @@
-import { StreamScope } from "@yext/studio-plugin";
+import { GetPathVal, StreamScope } from "@yext/studio-plugin";
 import { createContext } from "react";
 
 export interface AddPageData {
```
tmeyer2115, line 4: I'm also not wild about how we do `AddPageData`. It doesn't make a lot of sense to me that `AddPageData` doesn't contain all the info necessary to add a page. Before we added Layouts, it was missing name and getPathVal. Now it has those but is missing Layouts. I think this is something we should rework too, but not in this PR! The pattern was introduced before this PR.

packages/studio-ui/src/components/AddPageButton/BasicPageDataCollector.tsx
```diff
@@ -53,19 +54,27 @@ export default function BasicPageDataCollector({
...
-          setErrorMessage(err.message);
-          return false;
-        } else {
-          throw err;
-        }
+      if (isPagesJSRepo && !getPathValue) {
+        setErrorMessage("A getPath value is required.");
+        return false;
+      }
+      if (pages[data.pageName]) {
```
tmeyer2115, line 61: Right now, we're doing validation of `BasicPageData` in two different ways. We have a `PageDataValidator` class that we use above and we have these individual checks here. We should consolidate all our logic in one place. Can we add these new checks into `PageDataValidator`?

packages/studio-ui/src/components/AddPageButton/LayoutSelector.tsx
```diff
@@ -0,0 +1,68 @@
+import { ChangeEvent, useCallback, useState } from "react";
+import { FlowStepModalProps } from "./FlowStep";
+import useStudioStore from "../../store/useStudioStore";
+import DialogModal from "../common/DialogModal";
+import { selectCssClasses } from "../UnionPropInput";
+import { twMerge } from "tailwind-merge";
+
+export default function LayoutSelector({
```
tmeyer2115, line 8: Let's add some brief docs.

packages/studio-ui/src/components/AddPageButton/LayoutSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+  );
+
+  const selectClasses = twMerge(selectCssClasses, "w-full mb-6");
+  const body = (
+    <label>
+      <select className={selectClasses} onChange={handleChange}>
+        <option key="" value="">
+          No layout selected
+        </option>
+        {Object.entries(layouts).map(([layoutName]) => (
```
tmeyer2115, line 48: If you're only using the `layoutName` in the `map`, could we just do `Object.keys(layouts)` here?

packages/studio-ui/src/components/AddPageButton/LayoutSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+  );
+  const [selectedLayout, setSelectedLayout] = useState<string>("");
+  const [errorMessage, setErrorMessage] = useState<string>();
+
+  const onConfirm = useCallback(async () => {
+    const layoutState = layouts[selectedLayout];
+
+    try {
+      await handleConfirm(layoutState);
+      await handleClose();
```
tmeyer2115, line 24: Can either of these methods throw an exception? Do we need them in a try-catch?

packages/studio-ui/src/components/AddPageButton/AddPageContext.ts
alextaing: For sure, I agree its not super great. I tried to break out of this pattern, but the reason I had to keep along with it was because even if we add layoutState to `AddPageData` and set it once the user decides on a layout, the context does not update in time for it to be used by `createPage`.

packages/studio-ui/tests/components/AddPageFlow/AddPageButton.test.tsx
```diff
@@ -61,7 +99,7 @@ describe("PagesJS repo", () => {
     await userEvent.type(urlTextbox, url);
   }
 
-  it("closes modal when page type, name, and url are specified for static page", async () => {
+  it("closes modal after page type, name, and lauout for static page", async () => {
```
nmanu1, line 102: nit: lauout

packages/studio-ui/tests/components/AddPageFlow/AddPageButton.test.tsx
```diff
@@ -251,7 +309,7 @@ describe("errors", () => {
     await userEvent.click(addPageButton);
     const textbox = screen.getByRole("textbox");
     await userEvent.type(textbox, "universal");
-    const saveButton = screen.getByRole("button", { name: "Save" });
+    const saveButton = screen.getByRole("button", { name: "Next" });
```
nmanu1, line 312: nit: variable should be renamed here and below

packages/studio-ui/src/components/AddPageButton/BasicPageDataCollector.tsx
alextaing: I also see that we check for an existing page of the same name in a validation step within `createPage`.  Should we keep this in just in case, or can we delete it?  Since we check here, we shouldn't have a duplicate page name by the time we hit `createPage`.

packages/studio-ui/src/components/AddPageButton/AddPageContextProvider.tsx
```diff
@@ -16,9 +16,22 @@ export default function AddPageContextProvider(props: PropsWithChildren) {
...
+            return { ...state, streamScope };
+          }),
+        setPageName: (pageName: string) =>
+          setState((state) => {
+            return { ...state, pageName };
+          }),
+        setGetPathVal: (getPathVal?: GetPathVal) =>
+          setState((state) => {
+            return { ...state, getPathVal };
+          }),
```
oshi97, line 34: If all these are doing is setting values in the state object, I think it'd be simpler to expose setState instead of have 4 different actions

packages/studio-ui/src/components/AddPageButton/LayoutSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+      setSelectedLayout(e.target.value);
+    },
+    [setSelectedLayout]
+  );
+
+  const selectClasses = twMerge(selectCssClasses, "w-full mb-6");
+  const body = (
+    <label>
+      <select className={selectClasses} onChange={handleChange}>
+        <option key="" value="">
```
oshi97, line 45: Do we need the blank key and value?

packages/studio-ui/src/utils/PageDataValidator.ts
```diff
@@ -8,12 +10,22 @@ export interface ValidationResult {
...
+    isEntityPage,
+    isPagesJSRepo,
+    pages,
+  }: {
+    isEntityPage: boolean;
+    isPagesJSRepo: boolean;
+    pages?: PagesRecord;
+  }) {
+    this.isEntityPage = isEntityPage ?? false;
+    this.isPagesJSRepo = isPagesJSRepo ?? false;
```
nmanu1, line 27: since `isEntityPage` and `isPagesJSRepo` are required, we don't need this defaulting

packages/studio-ui/src/components/AddPageButton/LayoutSelector.tsx
alextaing: In our `handleConfirm`, we call create page which can throw errors if something goes wrong-- we then post the error on the modal instead of closing it.  Right now, we do checks for a `getPathValue` in PagesJS repos, and a check for invalid page names.

packages/studio-ui/src/components/AddPageButton/AddPageContext.ts
```diff
@@ -1,14 +1,18 @@
...
   isStatic: boolean;
   streamScope?: StreamScope;
+  pageName?: string;
```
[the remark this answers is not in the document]
alextaing: Yeah, that was the reason-- that makes sense, though, changed!

packages/studio-ui/tests/components/AddPageFlow/AddPageButton.test.tsx
```diff
@@ -213,7 +269,7 @@ describe("PagesJS repo", () => {
           entityFiles: ["mockLocalData.json"],
         },
       });
-      await waitFor(() => expect(screen.queryByText("Save")).toBeNull());
+      await waitFor(() => expect(screen.queryByText("Next")).toBeNull());
```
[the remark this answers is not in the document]
alextaing: good catch thanks

packages/studio-ui/src/utils/PageDataValidator.ts
alextaing: oops yeah I made them required to make it less confusing and to get rid of the defaulting-- just forgot to remove that

packages/studio-ui/src/components/AddPageButton/AddPageContextProvider.tsx
alextaing: oh, that's so true and makes a lot of sense.  We changed the setStates to function setStates to make sure that old values don't get overwritten if react calls them in batches.  Even through we have all state updates in once function, it is ok to keep the functional setState for safety?

oshi97: yeah that's fine!

packages/studio-ui/src/components/AddPageButton/BasicPageDataCollector.tsx
tmeyer2115: I think we can delete the duplicate check in `addPage`.

packages/studio-ui/src/components/AddPageButton/LayoutSelector.tsx
```diff
@@ -40,18 +45,24 @@ export default function LayoutSelector({
...
-    <label>
-      <select className={selectClasses} onChange={handleChange}>
-        <option key="" value="">
-          No layout selected
+    <select
+      className={selectClasses}
+      onChange={handleChange}
+      aria-label="Layout picker"
+    >
+      <option value="" aria-label="No layout selected">
```
nmanu1, line 53: I would have thought that `option` elements would automatically get labelled by their text content, so we wouldn't need to explicitly add an `aria-label`. Is that not the case?

e2e-tests/__screenshots__/darwin/add-entity-page.spec.ts/can-add-an-entity-page-2.png
alextaing, line 1: note: revisit screenshots

e2e-tests/tests/add-entity-page.spec.ts
```diff
@@ -12,11 +11,9 @@ studioTest("can add an entity page", async ({ page, studioPage }) => {
   const pageInTree = page.getByText("EntityPage");
   await expect(pageInTree).toHaveCount(0);
 
-  const streamScopeForm: StreamScopeForm = {
-    entityTypes: "entity1",
-    savedFilterIds: "entity2,entity3",
-  };
-  await studioPage.addEntityPage("EntityPage", streamScopeForm, "entity-page");
+  // TODO: Specify a stream scope once we can supply an API key for populating
+  // the store with account content
```
alextaing: note: todo

packages/studio-ui/src/components/AddPageButton/LayoutSelector.tsx
alextaing: Ooops well we don't need the aria labels so I removed them all as requested

packages/studio-ui/src/components/AddPageButton/LayoutSelector.tsx
alextaing: I thought I tried something like that earlier, but it actually does look like that works, changed!

packages/studio-ui/src/utils/PageDataValidator.ts
```diff
@@ -28,12 +40,20 @@ export default class PageDataValidator {
...
     const errorMessages: string[] = [];
-    if (pageData.pageName !== undefined)
-      errorMessages.push(...this.validatePageName(pageData.pageName));
-    if (pageData.url !== undefined)
-      errorMessages.push(
-        ...this.validateURLSlug(pageData.url, this.isEntityPage)
-      );
+    const { pageName, url } = pageData;
+    if (pageName !== undefined) {
+      errorMessages.push(...this.validatePageName(pageName));
```
tmeyer2115, line 45: Can we move the page name uniqueness check into `validatePageName`? Then we have only one place where we are doing page name validation.

packages/studio-ui/src/utils/PageDataValidator.ts
```diff
@@ -28,12 +40,20 @@ export default class PageDataValidator {
...
+      if (this.pages[pageName]) {
+        errorMessages.push(`Page name "${pageName}" is already used.`);
+      }
+    }
+
+    if (this.isPagesJSRepo && !url) {
```
tmeyer2115, line 51: Could we make a method called `validateURL` in this class? It would do this check and also invoke `this.validateURLSlug`. For code organization and readability purposes, I think having a consolidated `validateURL` method would be nice.

packages/studio-ui/src/components/AddPageButton/StreamScopeCollector.tsx
```diff
@@ -12,26 +12,28 @@ export default function StreamScopeCollector({
...
-  const [streamScope, setStreamScope] = useStreamScope();
+  const { state, actions } = useContext(AddPageContext);
+  const { streamScope } = state;
+  const { updateState } = actions;
 
   const onConfirm = useCallback(async () => {
-    if (streamScope === undefined) {
-      setStreamScope({});
-    }
+    updateState(streamScope ? { streamScope } : { streamScope: {} });
```
tmeyer2115, line 20: Is this `updateState` necessary? From a first glance, it seems like this method would be invoking `updateState` with the `streamScope` that's already been set in state. If the `streamScope` is undefined, `updateSelection` should have already put `{ streamScope: {} }` in the overall state.

packages/studio-ui/src/utils/PageDataValidator.ts
alextaing: that makes sense!

packages/studio-ui/src/components/AddPageButton/StreamScopeCollector.tsx
alextaing: I changed it to just set  `streamScope` to `{}` if a `streamScope` does not exist.  This is needed because if the user does not ever interact with the dropdown, the `updateSelection` will never be called, and `streamScope` will be `undefined` instead of `{}`!

tmeyer2115 (approved):

oshi97 (approved):

PR #377, opened 2023-09-20
Merge branch 'main' into feature/add-page-improvements

nmanu1:
Note: the size limit needed to be updated now that we are importing the default export from `react-select`


packages/studio-ui/src/components/EditStreamScopeModal.tsx
```diff
@@ -5,7 +5,8 @@ import StreamScopePicker, {
 } from "./StreamScopePicker";
 import { ResponseType, StreamScope } from "@yext/studio-plugin";
 import useStudioStore from "../store/useStudioStore";
-import { isEqual, sortBy } from "lodash";
+import isEqual from "lodash/isEqual";
+import sortBy from "lodash/sortBy";
```
alextaing, line 9: I think Oliver was mentioning in an [earlier PR](https://github.com/yext/studio/pull/374#discussion_r1329124742) that lodash per method packages are less efficient than just importing it wholesale according to [lodash](https://lodash.com/per-method-packages).

nmanu1: My understanding of what lodash is saying there is that they don't recommend installing the individual per-method packages (e.g. `lodash.isEqual`) and you should import from the overall `lodash` package like we're doing here (their example is `const throttle = require('lodash/throttle')`)

alextaing: ohh I see I think I misunderstood what they were saying, that makes sense

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -50,17 +50,22 @@ export default class ParsingOrchestrator {
...
     const pageFile = this.pageNameToPageFile[pageName];
     if (pageFile) {
       return pageFile;
     }
     return this.createPageFile(pageName);
-  }
+  };
 
-  private createPageFile(pageName: string) {
+  private createPageFile = (pageName: string) => {
```
alextaing, line 68: Just our of curiosity, what is the advantage of having this as an arrow function as opposed to a standard function declaration?

nmanu1: It has to do with `this` binding (see [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this)). Unlike standard functions, arrow functions don't bring their own `this` binding, so `this` retains the value from the calling function. Since we're passing `createPageFile` as a parameter to another function, we want it to retain `this` instead of creating a new binding for it, which would result in errors about accessing properties of undefined anywhere we do `this.blah` within the body of `createPageFile`. To do this, we can either change the syntax to an arrow function like we did here, or use the `bind` [method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#the_bind_method)

alextaing: oh ok, I see-- thanks for the explanation!

tmeyer2115 (approved):

PR #378, opened 2023-09-20
Update element icons

alextaing:
This PR updates element icons to the ones that product requested.  This PR also consolidates the logic we used to decide what icon to render. SVGs sourced from https://fontawesome.com.

image

J-SLAP=2940
TEST=manual


tmeyer2115:
LGTM once we regenerate the screenshots!

oshi97:
I'm okay not regenerating screenshots if it's not getting picked up automatically - I don't think it's worth spending time manually picking out all the ones that need to be updated. It would be better to spend that time updating the **way** we take those screenshots to be focused on smaller portions of the page. Eventually this will also let us decrease the minimum pixel change ratio

oshi97 (dismissed):

tmeyer2115 (approved):

PR #379, opened 2023-09-21
Live Preview Button Opens the Corresponding Static/Entity Page.

tmeyer2115:
This PR changes the behavior of the "Live Preview" button. Previously, this button would open the PagesJS Landing Page in a new tab. Now, it opens a PagesJS-powered preview of the active Page in Studio. If the active Page is an Entity Page, the preview uses the currently selected Entity in Studio. There are a couple scenarios in which the Live Preview button will be disabled: 

1. When we cannot properly determine the URL of active Page. 
2. If there is no active Page currently in Studio. 

In a later PR, I will add a Tooltip that appears when the button is disabled. It will explain why it's disabled. In this PR, I also removed the "Open Live Preview" Acceptance test. I don't think that test adds much value any more. We have extensive unit test coverage for the Live Preview button. This coverage ensures that, when enabled, `window.open` is called with the correct URL. We can assume that `window.open` works correctly. We already have various screenshots that capture the button (in both the enabled and disabled state). So, we've already got visual regression coverage. 

TEST=auto,manual

Ensured all the new unit tests passed. Using the Test Site, I verified that the proper Preview tab opened for both a Static and Entity page.

### DIFF
--- a/packages/studio-ui/tests/components/OpenLivePreviewButton.test.tsx
+++ b/packages/studio-ui/tests/components/OpenLivePreviewButton.test.tsx
@@ -22,8 +22,8 @@ describe("button is disabled properly", () => {
     mockStore({
       pages: {
         getActivePageState: () => undefined,
-        getActiveEntityData: () => undefined
-      }
+        getActiveEntityData: () => undefined,
+      },
     });
 
     render(<OpenLivePreviewButton />);


packages/studio-ui/src/components/OpenLivePreviewButton.tsx
```diff
@@ -1,19 +1,81 @@
...
+  if (isEntityPage) {
+    return entityData?.id
+      ? Result.ok(`${PAGES_JS_LANDING_PAGE}/${pageName}/${entityData.id}`)
+      : Result.err(new Error("Bad"));
+  }
+
+  const getPathVal = pagesJSState.getPathValue;
+  return !!getPathVal
+    ? Result.ok(`${PAGES_JS_LANDING_PAGE}/${getPathVal.value}`)
+    : Result.err(new Error("Bad"));
```
alextaing, line 80: Should we specify a more detailed error?

packages/studio-ui/src/components/OpenLivePreviewButton.tsx
```diff
@@ -1,19 +1,81 @@
...
+  );
+
   return (
     <div className="relative inline-block">
-      <a
-        className="rounded-md text-white bg-blue-600 px-2 py-1 flex items-center gap-x-2 shadow-md hover:bg-blue-700 hover:shadow-lg"
-        href="http://localhost:5173/"
-        target="_blank"
-        rel="noreferrer"
+      <button
```
alextaing, line 54: Just a thought-- but if a user has a non-pagesJS repo, should we just not display the button at all since it will never become enabled for that repo?

packages/studio-ui/src/components/OpenLivePreviewButton.tsx
tmeyer2115: Yep! Totally forgot, haha. Added more detailed messages.

alextaing (approved):
lgtm! 🥳

packages/studio-ui/src/components/OpenLivePreviewButton.tsx
```diff
@@ -1,19 +1,81 @@
+import { PagesJsState } from "@yext/studio-plugin";
+import useStudioStore from "../store/useStudioStore";
+import classNames from "classnames";
+import { Result } from "true-myth";
+import { useCallback } from "react";
+
+// The URL of the Landing Page of the PagesJS Dev Server. Port 5173 is hardcoded for now as it will be the PagesJS Dev port in most cases.
+export const PAGES_JS_LANDING_PAGE = "http://localhost:5173";
```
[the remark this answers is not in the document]
tmeyer2115: I will make an item to address!

PR #375, opened 2023-09-15
Add Layouts via AddElementMenu

oshi97:
This PR implements adding layouts via the AddElementMenu/Button.

J=SLAP-2924
TEST=manual,auto

I can add layouts multiple times, set the newly added components as active, and save to file
updated AddElementMenu tests

PR #369, opened 2023-09-11
Remove repeaters and modules

alextaing:
This PR removes repeaters and modules from Studio 👋 

J=SLAP-2928
TEST=auto

PR #382, opened 2023-09-22
Version 0.23.0.

tmeyer2115:
## Features
- We've introduced the concept of Page Layouts (or Layouts) to Studio. A Page Layout is a pre-prescribed set of Components that you can apply to a Page. By default, Layouts live in the `src/layouts` directory. In the UI, a Layout can be applied via the "Insert" button. (#370, #367, #375)
- There is now improved UX for creating or editing an Entity Page's Stream Scope. Instead of being raw inputs, the fields for Entity Ids, Entity Types, and Saved Filters are now drop-downs with the relevant options populated from the account. (#380)

## Changes
- Repeaters and Modules have been officially deprecated and removed from Studio. (#369)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24460,11 +24460,11 @@
     },
     "packages/studio": {
       "name": "@yext/studio",
-      "version": "0.22.0",
+      "version": "0.23.0",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


alextaing (approved):
LGTM 🔥

PR #383, opened 2023-09-22
v0.23.1

oshi97:
### Fixes

* Removes the tsconfig extends from the studio tsconfig.
* Adds virtual modules to optimizeDeps.exclude in studio's vite config.


oshi97:
I created an item to investigate why we need to specify these deps as excluded
https://yexttest.atlassian.net/browse/SLAP-2943

packages/studio-ui/package.json
```diff
@@ -62,7 +62,7 @@
     "@types/react": "^18.0.26",
     "@types/react-dom": "^18.0.10",
     "@types/react-modal": "3.13.1",
-    "@yext/studio-plugin": "0.23.0",
+    "@yext/studio-plugin": "*",
```
oshi97, line 65: these can be *

alextaing (approved):

PR #385, opened 2023-09-25
Fix Deploy button

alextaing:
The Deploy button had unexpected behavior since Git data (`canPush`) was not properly updated using HMR. This PR makes sure that on Studio save, updated Git data is sent to Studio.  A new type of HMR message was added between studio which handles exclusively Git data, as to not pollute the current HMR message type which is purely studio data from the studio data virtual module.

J=SLAP-2944
TEST=manual

🚧  Tests are currently being added 🚧


packages/studio-ui/src/store/useStudioStore.ts
```diff
@@ -49,6 +50,7 @@ const useStudioStore = create<StudioStore>()(
         ),
         studioConfig: lens(createStudioConfigSlice),
         studioEnvData: lens(createStudioEnvDataSlice),
+        studioGitData: lens(createStudioGitDataSlice),
```
oshi97, line 53: I think just gitData is fine

oshi97: arguably studioEnvData can just be envData too but whatever

packages/studio-plugin/src/git/reloadGitData.ts
```diff
@@ -14,4 +15,11 @@ export default async function reloadGitData(
     throw new Error("Expected a GitData module.");
   }
   server.moduleGraph.invalidateModule(gitDataModule);
```
oshi97, line 17: invalidateModule was here to get hot reloading for free. I think we can delete it now

packages/studio-ui/src/store/useStudioStore.ts
alextaing: renamed both!

packages/studio-plugin/src/git/reloadGitData.ts
alextaing: I notice we still do the invalidation for studio data too even though we also send our own custom HMR update.  Should we remove it there as well?

oshi97: actually I think we still might need it for when you refresh the page, not sure

oshi97 (dismissed):

tmeyer2115 (approved):

PR #386, opened 2023-09-25
Highlight current Viewport in `ViewportMenu` Component.

tmeyer2115:
This PR makes sure the currently applied Viewport is highlighted in the `ViewportMenu` drop-down. Previously, there was no highlighting. So, people could not remember what Viewport they'd applied. Note that we do not highlight the "Reset Viewport" option. I also added a visual regression test to ensure the correct highlighting behavior.

J=SLAP-2946
TEST=auto, manual


e2e-tests/__screenshots__/win32/viewport-menu.spec.ts/can-set-and-reset-Viewport-2.png
alextaing, line 1: Might we want to choose a smaller viewport for screenshot testing?  I think the `Galaxy Z Flip5 Folded` is almost identical to the normal reset viewport size.  My only concern is that the difference might be small such that any discrepancies won't register if we have issues in the future.

other than that, lgtm!

alextaing (approved):
🔥

PR #388, opened 2023-09-26
Add LoadingOverlay for Component And Bundle Loading

oshi97:
This PR adds the LoadingOverlay components, which will provide a loading screen while user components and
the rest of the application are being loaded over the network.

In the future we can add additional information to this screen regarding hitting the management api and loading
in localData files, if necessary.

J=SLAP-2895
TEST=manual


packages/studio-ui/src/utils/getFunctionComponent.ts
```diff
@@ -1,15 +1,7 @@
-import { FunctionComponent } from "react";
 import { ImportType } from "../store/models/ImportType";
 
 export default function getFunctionComponent(
-  importedValue: Record<string, unknown>,
-  name: string
+  importedValue: Record<string, unknown>
 ): ImportType | undefined {
-  if (typeof importedValue[name] === "function") {
```
oshi97, line 8: We don't support components that are not default exported, I think we probably used to which is why we had this logic but haven't checked. Possibly is related to the external components support that Tristan did that we had to cut

packages/studio-ui/src/AppWithLazyLoading.tsx
```diff
@@ -0,0 +1,75 @@
...
+          {renderComponentLoadingProgress(
+            loadedCount,
+            totalCount,
+            componentsLoaded
+          )}
+          {renderBundleMessage(appLoaded)}
+        </>
+      }
+    >
+      <Suspense>
```
alextaing, line 38: if we have the loading overlay, do we still need the suspense?  Or does it do something else other than display a fallback UI?

alextaing: is it just so everything in the app renders at once?

packages/studio-ui/.size-limit.cjs
```diff
@@ -1,6 +1,6 @@
 module.exports = [
   {
-    path: "lib/src/index.js",
+    path: "lib/**/*.js",
```
alextaing, line 3: more of a question for me, but what is the reason behind this change?

packages/studio-ui/src/utils/loadComponents.tsx
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Load all of the user's components into the store.
+ */
+export default function loadComponents(): Promise<void>[] {
+  const { UUIDToFileMetadata, setImportedComponent, UUIDToImportedComponent } =
+    useStudioStore.getState().fileMetadatas;
+
+  const componentImportPromises = Object.values(UUIDToFileMetadata).map(
+    async (fileMetadata) => {
+      if (UUIDToImportedComponent.hasOwnProperty(fileMetadata.metadataUUID)) {
```
alextaing, line 14: also for my edification, would something like `if (fileMetadata.metadataUUID in UUIDToImportedComponent)` be more readable, or is there a functional difference for this case?

alextaing: its OK, found an article: https://masteringjs.io/tutorials/fundamentals/hasownproperty

packages/studio-ui/.size-limit.cjs
oshi97: Adding a dynamic import to the bundle adds additional js files to studio-ui/lib that don't live under the src folder

packages/studio-ui/src/utils/loadComponents.tsx
oshi97: oh yeah distinction is small and 99% of the time unimportant

e2e-tests/__screenshots__/darwin/add-element.spec.ts/can-add-a-Footer-component-1.png
alextaing, line 1: Not related to the changes in this PR, but I think at some point we lost our local data or something b/c the hours component looks like it broke.  Should we look into adding this back at some point?  It also looks like our styling got messed up too for the BusinessInfo container.

alextaing: other than that, lgtm!

alextaing (dismissed):

e2e-tests/__screenshots__/darwin/add-element.spec.ts/can-add-a-Footer-component-1.png
oshi97: I regenerated the local data!

alextaing: should we regen screenshots, or do you want to just do that as a part of the larger screenshot item?

packages/studio-ui/src/AppWithLazyLoading.tsx
```diff
@@ -0,0 +1,75 @@
...
+      </div>
+    </>
+  );
+}
+
+function renderBundleMessage(appLoaded: boolean) {
+  const className = classNames("text-sky-600 mt-4", {
+    "animate-pulse": !appLoaded,
+  });
+  const msg = appLoaded ? "JS bundle loaded!" : "... loading JS bundle ...";
```
tmeyer2115, line 72: Small nit: Can we come up with a more friendly message for non-technical users? They won't know what a JS bundle is.

e2e-tests/__screenshots__/darwin/add-element.spec.ts/can-add-a-Footer-component-1.png
oshi97: I don't think it's worth bothering with if playwright can't even detect that the screenshot has changed

packages/studio-ui/src/AppWithLazyLoading.tsx
oshi97: updated!

alextaing (approved):
⚡

PR #395, opened 2023-10-04
Support inheritance when parsing interfaces

nmanu1:
This PR adds support for inherited properties when parsing interfaces such as component prop interfaces or site settings. Something like this is now supported:
```ts
import { CtaData } from "@yext/search-ui-react";

interface BaseProps {
  title: string;
}

interface ComponentProps extends BaseProps, CtaData {
  description: string;
}
```
If we're not able to resolve the type that's being extended to an object type, we throw an error. This would include extending a non-object type like `HexColor` (which is invalid TS), extending a built-in type like `Record`, etc. We also aren't adding any additional support for type unions or intersections, beyond our current support for string unions, since TS does more complicated merging when properties are shared in those cases. For interfaces that inherit properties, if a property is shared between the inherited type and the new type, the new property's type is required by TS to be a subset of the inherited property's type, and the new property's type is used in the final, generated type. In other words, only type narrowing is allowed.

J=SLAP-2938
TEST=auto, manual

In the test-site, see that prop editors for inherited props show up as expected in the UI.


packages/studio-plugin/src/parsers/helpers/TypeNodeParsingHelper.ts
```diff
@@ -137,12 +137,49 @@ export default class TypeNodeParsingHelper {
...
+        inheritedType.getExpression()
+      );
+      const parsedType = parseTypeReference(typeName);
+      if (parsedType?.kind === ParsedTypeKind.Object) {
+        return {
+          ...inheritedShape,
+          ...parsedType.type,
+        };
+      } else {
+        throw new Error(
```
tmeyer2115, line 167: In the case of parsing a Component's Prop Interface, if this `Error` is thrown, do we gracefully handle? That is, do we put the Component in the Error State in the UI?

nmanu1: Yes, as with other parsing errors, this puts the component in an error state. This means in the UI we remove the component from the Add Element list and display any page's usages of the component in red with a tooltip (both in the component tree and the preview).

Screenshot 2023-10-05 at 9 43 33 AM

PR #396, opened 2023-10-05
Create `ComponentTreeSection` and Take More Scoped Screenshots. 

tmeyer2115:
This PR creates the `ComponentTreeSection`. This Section is a Playwright wrapper around the Component Tree functionality. As part of this PR, I updated the `remove-element.spec.ts` test to take scoped screenshots of the actual Component Tree when removing a Component. 

J=SLAP-2922
TEST=auto


e2e-tests/tests/infra/appSections/ComponentTreeSection.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.removeElementButton = page.getByRole("button", {
+      name: "Remove Element",
+    });
+  }
+
+  getElement(name: string, index = 0): Locator {
+    return this.componentTree.getByText(name).nth(index);
+  }
+
+  async setActiveElement(name: string, index = 0) {
```
nmanu1, line 22: Maybe I missed it, but do we actually pass in an `index` anywhere?

e2e-tests/tests/remove-element.spec.ts
```diff
@@ -10,11 +10,17 @@ const expectedPage = fs.readFileSync(
...
   await expect(childPreviews).toHaveCount(1);
-  await studioPage.removeElement("Container");
+  await componentTreeSection.takeScreenshot();
+  await componentTreeSection.removeElement("Container");
+  await componentTreeSection.takeScreenshot();
   await expect(containerPreviews).toHaveCount(0);
   await expect(childPreviews).toHaveCount(0);
+
+  // TODO: Scope this to Screenshot of Preview pane.
   await studioPage.takePageScreenshotAfterImgRender();
```
nmanu1, line 24: Are we planning on changing most of our currently un-scoped screenshots to be scoped to the preview panel?

e2e-tests/tests/infra/appSections/ComponentTreeSection.ts
tmeyer2115: I don't think so. I kept it when I cut over the method from `StudioPlaywrightPage`. I could see us potentially using this in the future and it's not adding much code/complexity.

e2e-tests/tests/remove-element.spec.ts
tmeyer2115: I think a number of un-scoped screenshots will become scoped to the preview panel, yes. There will also be a few situations in which we replace the un-scoped screenshot with the screenshot of one of the other Section models.

nmanu1 (approved):

PR #397, opened 2023-10-05
Inject User CSS Into Preview Panel

alextaing:
This PR injects user specified CSS into the preview panel.  

**Components**
Studio-plugin generates a dependency graph for each component to trace back what CSS each component requires.  The CSS files for each component are stored in an array called `cssImports` within `ComponentMetadata`.  When the iframe is rendered, we look at the component tree, and for each component we dynamically import the required CSS files.  When the component tree or active page changes, we clear the CSS from the iframe and re-import CSS to ensure that the CSS in the iframe reflects what components are in the current active page.

**Templates**
Before, we took the relative file paths of css file imports in templates and saved them in state.  Now, we take the absolute file paths and store them in state.  This way we can dynamically import the CSS when needed.

Another overall issue was that since we use Vite to dynamically import our CSS files, a style tag was automatically placed in Studio.  This was bad since we don't want user styling to affect studio.  To get around this, a plugin was written that ensures that when we import a CSS file, we add the `?inline` tag onto it which tells Vite not to add a style tag for us.

Additionally, in the past the Tailwind that we use for Studio also affected the iframe. This means that if the user does not have their own repos set up properly for Tailwind, the Studio preview would be inaccurate since they don't actually have Tailwind.  In doing this, we realized that our test site and e2e test site were not configured for Tailwind, so this was also done in this PR.

J-SLAP-2912
TEST=manual, auto


packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,47 @@
...
+          upath.dirname(importer),
+          addQueryParameter(id, "inline")
+        );
+      }
+    },
+  };
+}
+
+function isImportedByStudio(id: string) {
+  return !!(id.match(/.*@yext\/studio.*/) || id.match(/.*packages\/studio*/));
```
alextaing, line 34: I'm not a huge fan of the way this is currently done-- it's just a regex checking the path of what imported the file.  I'm not sure what we can assume.

packages/studio-ui/src/components/IFramePortal.tsx
```diff
@@ -42,21 +44,62 @@ export default function IFramePortal(
...
+      return;
+    }
+    componentTree?.forEach((component) => {
+      const metadataUUID = component.metadataUUID;
+      if (!metadataUUID) {
+        return;
       }
+      const cssImports = getComponentMetadata(metadataUUID).cssImports;
+      cssImports?.forEach((cssFilepath) => {
+        void dynamicImportFromBrowser(cssFilepath + "?inline").then(
```
alextaing, line 67: Even though we add the `inline` query param in the plugin, we need to have this `inline` query param here as well or else the function will not properly return the contents of the dynamic import (i.e. the import will happen, but it won't be accessible in the `then` since it seems unable to map back to the data if the id has changed).  I think it is just an idiosyncrasy of the dynamic import function.

PR #398, opened 2023-10-06
Fix imports for commonly-named components

nmanu1:
This PR fixes imports for commonly-named components like `Container` and `Text` by manually writing their imports to the file before calling `ts-morph`s `fixMissingImports` function. Previously, this function would often add an incorrect import in the case of `Container` and not add any import in the case of `Text`, which is also a built-in interface in React.

J=SLAP-2587
TEST=auto, manual

Added `postcss` to the test site and saw that without these changes, `Container` was imported from `postcss` (which also has a named export called `Container`), and that with these changes, `Container` is correctly imported from `src/components`. Also saw that the behavior for `Text` was fixed (i.e. an import was actually added for a `Text` component rather than it trying to use the React type).

### DIFF
--- a/packages/studio-plugin/src/sourcefiles/PageFile.ts
+++ b/packages/studio-plugin/src/sourcefiles/PageFile.ts
@@ -1,6 +1,6 @@
 import { ArrowFunction, FunctionDeclaration, Project } from "ts-morph";
 import { Result } from "true-myth";
-import { PageState } from "../types";
+import { FileMetadata, PageState } from "../types";
 import TemplateConfigWriter from "../writers/TemplateConfigWriter";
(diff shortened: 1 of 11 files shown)


nmanu1:
> Since the impetus of the item was the incorrect Container import, could you add `postcss` to the test site? I'd like to verify that we've fixed the bug for `Container`. We don't need to check in those changes to the test site, I'd just like us to verify it manually.

Added `postcss` to the test site and confirmed that without the changes in this PR, `Container` was imported from `postcss` and with the changes, `Container` is correctly imported from `src/components`!

tmeyer2115 (approved):

PR #400, opened 2023-10-09
Remove redundant snapshots from 'Add Page Using a Layout' test.

tmeyer2115:
This PR updates `add-page-with-layout.spec.ts` to use the new `AddPageSection`. In addition, I've removed redundant snapshots. We have snapshots of the modal flow for adding a static page w/out a Layout. So, here, we just need a snapshot of the Layout Modal. We also don't need a snapshot of the newly created page pre-save. We test that the "Save" button is enabled by clicking it. And we have visual regression tests already for the enabled "Save" button. 

J=SLAP-2922
TEST=auto


e2e-tests/tests/add-page-with-layout.spec.ts
```diff
@@ -9,12 +9,15 @@ const expectedPage = fs.readFileSync(
...
   await expect(pageInTree).toHaveCount(0);
-  await studioPage.addStaticPage(pageName, "index.html", "LocationLayout");
+
+  const addPageSection = studioPage.addPageSection;
+  await addPageSection.selectPageType(false);
+  await addPageSection.enterBasicPageData(pageName, "index.html");
+  await addPageSection.selectLayout("LocationLayout", true);
+
   await expect(pageInTree).toHaveCount(1);
-  await studioPage.takePageScreenshotAfterImgRender();
```
nmanu1, line 17: I noticed pretty much all our tests have a screenshot from before we click the Save button and a screenshot for after. Are we planning on removing this "before" screenshot from all of the tests?

e2e-tests/__screenshots__/add-page-with-layout.spec.ts/can-add-a-page-using-a-layout-1.png
nmanu1, line 1: Now that we have more scoped screenshots, do we also want to add ones for other states of this modal, for example, a screenshot for the initial state before a layout is selected and a screenshot for what the dropdown of layouts looks like?

e2e-tests/tests/add-page-with-layout.spec.ts
tmeyer2115: Yep, that's the plan.

e2e-tests/__screenshots__/add-page-with-layout.spec.ts/can-add-a-page-using-a-layout-1.png
tmeyer2115: We now have screenshots for the initial state and when an option is selected. Getting a screenshot of the expanded dropdown is proving tricky. I added a TODO though. Prior to this PR, we just had a screenshots of the initial state and once an option is selected. So, we're not losing regression test coverage with this PR.

e2e-tests/tests/infra/appSections/AddPageSection.ts
```diff
@@ -1,11 +1,11 @@
-import { Page, expect } from "@playwright/test";
+import { Locator, Page, expect } from "@playwright/test";
 
 /**
  * A Playwright wrapper around the Add Page flow.
  */
 export default class AddPageSection {
   private addPageButton;
```
nmanu1, line 7: nit: can we add a type for this too?

nmanu1 (approved):

PR #401, opened 2023-10-09
Use `AddPageSection` in Playwright When Adding an Entity Page. 

tmeyer2115:
This PR updates the `can-add-entity-page` test to use the `AddPageSection`. With that, all logic for adding a Static/Entity Page can finally be removed from `StudioPlaywrightPage`. I cut down the number of snapshots taken by this test significantly. It only takes snapshots pertaining to Scope Selection (the one part of the Add Page Flow we still need regression coverage for) and once the new Entity Page is saved. 

I added a TODO. To further improve our visual regression test coverage, we should have the Scope Selection Modal actually be populated with data that we can expand. But, that's outside the scope of this PR. 

J=SLAP-2922
TEST=auto


e2e-tests/tests/infra/appSections/AddPageSection.ts
```diff
@@ -1,4 +1,9 @@
 import { Locator, Page, expect } from "@playwright/test";
+import { StreamScope } from "@yext/studio-plugin";
+
+type StreamScopeForm = {
+  [key in keyof StreamScope]: string;
```
alextaing, line 5: I guess we don't use this quite yet, but would we want the values to be `string[]`?

tmeyer2115: Updated!

alextaing (approved):

e2e-tests/tests/add-entity-page.spec.ts
```diff
@@ -11,11 +11,15 @@ studioTest("can add an entity page", async ({ page, studioPage }) => {
   const pageInTree = page.getByText("EntityPage");
   await expect(pageInTree).toHaveCount(0);
 
+  const addPageSection = studioPage.addPageSection;
+  await addPageSection.selectPageType(true);
+  await addPageSection.enterStreamScope({}, true);
+  await addPageSection.enterBasicPageData("EntityPage", "entity-page");
```
nmanu1, line 17: The basic page data modal is a little different between static and entity pages. For entity pages, the URL slug field should be populated with `[[slug]]`. Do we want to add a screenshot for that?

e2e-tests/tests/add-entity-page.spec.ts
```diff
@@ -11,11 +11,15 @@ studioTest("can add an entity page", async ({ page, studioPage }) => {
...
+  await addPageSection.selectLayout();
+
   // TODO: Specify a stream scope once we can supply an API key for populating
   // the store with account content
```
nmanu1, line 21: nit: Can this TODO be moved above line 16 where we currently pass in an empty stream scope? Looks like this TODO was missed when we did the API injection work. Can we make an item to address it and the other outstanding TODO from that work (in `useStudioStore`)?

PR #402, opened 2023-10-10
Add Component CSS imports to state

alextaing:
This PR adds a list in `ComponentMetadata` of the file paths to the CSS files that each component imports.  In `ComponentFile`, a dependency tree is used to find all transitive CSS for each component. The `dependency-tree` library allows for optimization by passing in already parsed dependency trees, so component tree is formed in Parsing Orchestrator so that previous dependencies can be accessed.

J=SLAP-2912
TEST=auto,manual


packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -194,7 +212,11 @@ export default class ParsingOrchestrator {
     });
 
     if (absPath.startsWith(this.paths.components)) {
-      const componentFile = new ComponentFile(absPath, this.project);
+      const componentFile = new ComponentFile(
+        absPath,
+        this.project,
+        this.getComponentDependencyTree(absPath)
```
tmeyer2115, line 218: I think the way in which we create the Dependency Tree in ParsingOrchestrator is a bit convoluted. Here, we're trying to create the `ComponentFile` without the requisite information (the Component's Dependency tree). So we create it on the fly and update a global variable (`this.dependencyTree`). There are two distinct places where we invoke `updateDependencyTree`. 

As an alternative, could we first create the full `dependencyTree` and then run `initFilepathToFileMetadata`? Or would that cause a performance hit?

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -180,6 +183,21 @@ export default class ParsingOrchestrator {
...
 
+  private updateDependencyTree(absPath) {
+    this.dependencyTree[absPath] = dependencyTree({
+      filename: absPath,
+      directory: upath.dirname(absPath),
+      visited: this.dependencyTree,
+    });
+  }
+
+  private getComponentDependencyTree(absFilepath: string) {
```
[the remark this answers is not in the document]
alextaing: oh yes, changed!

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
alextaing: We currently initialize the `dependencyTree` as a part of `initFilepathToFileMetadata`, and I can move `updateDependencyTree` before we `getFileMetadata` so it should have the requisite information by the time it gets to it when we first initialize!  However, we probably won't be able to avoid using some sort of "on the fly" tree updating, since we support HMR.  This is important in the case when a developer adds a new component file and HMR makes it appear instantly in Studio.  Since we won't have that file's dependency tree, we'll need to generate it on the spot.

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
```diff
@@ -15,12 +16,14 @@ import { Result } from "true-myth";
 export default class ComponentFile {
   private studioSourceFileParser: StudioSourceFileParser;
   private fileMetadataParser: FileMetadataParser;
+  private dependencyTree: Tree;
 
-  constructor(filepath: string, project: Project) {
+  constructor(filepath: string, project: Project, dependencyTree: Tree) {
```
nmanu1, line 21: nit: you can just use the `private dependencyTree: Tree` shorthand and it'll create a `private` variable for the class instantiated to the value passed into the constructor

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
```diff
@@ -42,12 +45,27 @@ export default class ComponentFile {
...
       ...(acceptsChildren ? { acceptsChildren } : {}),
       filepath,
+      cssImports,
     };
   };
 }
+
+function getCssFilesFromDependencyTree(dependencyTree: Tree): string[] {
+  let cssFiles: Set<string> = new Set();
+  Object.entries(dependencyTree).forEach(([filename, subDependencyTree]) => {
```
nmanu1, line 61: if we use `reduce` here instead of `forEach`, I think that should simplify it a little and we should be able to remove the `let`

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
alextaing: done!

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -159,25 +162,56 @@ export default class ParsingOrchestrator {
...
+
+  private updateDependencyTree(absPath: string) {
+    this.dependencyTree[absPath] = dependencyTree({
+      filename: absPath,
+      directory: upath.dirname(absPath),
+      visited: this.dependencyTree,
+    });
+  }
+
+  private getOrCreateDependencyTree(absFilepath: string): Tree {
```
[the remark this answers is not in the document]
alextaing: renamed!

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
alextaing: I gave it a shot, and I think since I'm using a `Set` to avoid repetition, it won't work since its not iterable in the same way arrays are, interestingly.

I could do some stuff with arrays, but I think that would make the code a little messy.  Are you OK with how it is now?

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
alextaing: Update to this-- we update dependency tree in `getFileMetadata`, since that is called in the `filepathToFileMetadata` initializer, so this initializes the dependency tree in the beginning.  Another reason we want to call this here is because if the user were to update any files with new CSS, we would want a new, updated dependency tree to make sure that their CSS imports are up to date.   This also just happens simplifies the code!

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
alextaing: Ohh I see-- yeah that works.  I did the same thing, except instead of ` new Set<string>()` as the initial value, I did ` {} as Set<string>` which broke the site.  Changed!

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -208,6 +215,15 @@ export default class ParsingOrchestrator {
     );
   };
 
+  private updateFilepathToDependencyTree(absPath: string) {
+    delete this.filepathToDependencyTree[absPath];
```
nmanu1, line 219: do we need this `delete`? we reassign below anyways

alextaing: yes! we use `filepathToDependencyTree` to as "cached" info to speed up updating the dependency tree, so we delete it so we don't just update the tree with cached info.

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
```diff
@@ -42,12 +47,29 @@ export default class ComponentFile {
...
+        cssFiles.add(filename);
+      }
+      return new Set([
+        ...cssFiles,
+        ...getCssFilesFromDependencyTree(subDependencyTree),
+      ]);
+    },
+    new Set<string>()
+  );
+  return Array.from(cssFiles);
```
nmanu1, line 74: nit: I slightly prefer the spread syntax here instead of `Array.from`. I think it's supposed to be a little faster

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
nmanu1, line 219: should this `delete` live in `reloadFile`? in the case where a component file is deleted (HMR), I think this code wouldn't be reached?

packages/studio-plugin/tests/__fixtures__/ParsingOrchestrator/src/components/Card.tsx
```diff
@@ -1,3 +1,4 @@
+import "../index.css";
```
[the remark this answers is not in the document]
alextaing: I can add one in a later PR!  The CSS is just in state and not on the page yet

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -208,6 +216,14 @@ export default class ParsingOrchestrator {
     );
   };
 
+  private updateFilepathToDependencyTree(absPath: string) {
```
nmanu1, line 219: do we need to think about calling this function as result of HMR triggered by a change to a dependency of a component file?

alextaing: decided to hold off on this until we support general HMR for this case, not just CSS!

packages/studio-plugin/src/types/ComponentMetadata.ts
```diff
@@ -9,4 +9,5 @@ export type ComponentMetadata = {
   initialProps?: PropValues;
   propShape?: PropShape;
   acceptsChildren?: boolean;
+  cssImports?: string[];
```
nmanu1, line 12: do we want this field to be required like it is in PageState? it looks like it'll always be specified, even if it's just an empty array

packages/studio-plugin/tests/__fixtures__/ParsingOrchestrator/src/components/Card.tsx
nmanu1, line 1: can we test a nested CSS import as well to make sure the dependencies are being traced correctly?

alextaing: done!

nmanu1 (approved):
🔥

PR #403, opened 2023-10-10
Add an Info Icon for Props with a Tooltip.

tmeyer2115:
This PR adds an Info icon to the right of the Prop Name when the Prop has a Tooltip. The anchor of the Tooltip is now the Info Icon. To do this, I created a new common Component called TooltipIcon representing the combination of an `Icon` and a `Tooltip`. I added visual regression coverage for Prop Tooltips as well, we were missing it. 

J=SLAP-2952
TEST=auto


packages/studio-ui/src/components/PropEditor.tsx
```diff
@@ -48,9 +49,18 @@ export default function PropEditor({
...
-        </label>
+        <div className="flex flex-row">
+          <label className="pb-1" htmlFor={inputId}>
+            {propName}
+          </label>
+          {tooltip && (
+            <Info
+              id={labelTooltipId}
+              className="ml-3 pb-1"
+              data-testid="prop-tooltip"
```
nmanu1, line 60: Could we use an aria label instead of a data attribute here?

packages/studio-ui/src/components/PropEditor.tsx
```diff
@@ -48,9 +49,18 @@ export default function PropEditor({
...
-        <label className="pb-1" id={labelTooltipId} htmlFor={inputId}>
-          {propName}
-        </label>
+        <div className="flex flex-row">
+          <label className="pb-1" htmlFor={inputId}>
+            {propName}
+          </label>
+          {tooltip && (
+            <Info
+              id={labelTooltipId}
```
nmanu1, line 58: nit: can the variable name and value be updated now that the ID is no longer for a label element?

packages/studio-ui/src/components/PropEditor.tsx
```diff
@@ -48,9 +49,18 @@ export default function PropEditor({
...
         {isNested && renderBranchUI()}
-        <label className="pb-1" id={labelTooltipId} htmlFor={inputId}>
-          {propName}
-        </label>
+        <div className="flex flex-row">
+          <label className="pb-1" htmlFor={inputId}>
+            {propName}
+          </label>
+          {tooltip && (
+            <Info
```
nmanu1, line 57: What do you think of separating out this Info icon and the actual tooltip into a common component? That would make it easier to reuse if we decide we want this icon for tooltips in other places in the Studio UI

packages/studio-ui/src/components/PropEditor.tsx
tmeyer2115: If it's okay, I'd like to stick with the Playwright `testid` for now. Working with the Tooltip, I've come up with some more general WCAG questions for Product. I think those would influence what we want to do ARIA-wise.

alextaing (approved):
unless nidhi has more comments, lgtm!

PR #404, opened 2023-10-10
Absolute Path Page CSS Imports

alextaing:
This PR modifies how page CSS is parsed.  Instead of storing relative CSS file paths in `cssImports` for `pageState`, we now store absolute CSS paths or paths from a node module package.  This is so we can easily dynamically import them in `studio-ui`.  Since we do this, we have to make sure that when we write back the CSS import, it is transformed back into the relative path.

J=SLAP-2912
TEST=auto,manual


packages/studio-plugin/src/parsers/StudioSourceFileParser.ts
```diff
@@ -101,11 +101,21 @@ export default class StudioSourceFileParser {
 
   parseCssImports(): string[] {
     const cssImports: string[] = [];
+    const getAbsolutePath = (source: string) => {
+      if (source.startsWith(".")) {
+        const absPath = upath.join(
```
nmanu1, line 106: is there a way to use `upath.resolve` instead of `join` here? then we can likely remove the check on line 105 and I think it would make it a little more clear that we're definitely returning an absolute path

packages/studio-plugin/tests/sourcefiles/LayoutFile.test.ts
```diff
@@ -87,7 +87,7 @@ describe("getLayoutState", () => {
   it("correctly parses CSS imports", () => {
     const result = getLayoutState("BasicLayout");
     expect(result.value.cssImports).toEqual([
-      "./index.css",
+      expect.stringContaining("index.css"),
```
nmanu1, line 90: nit: in this test and the one for PageFile, can we check the exact path that's computed since that's the key part of this change set?

packages/studio-plugin/src/parsers/StudioSourceFileParser.ts
alextaing: used path.resolve!

packages/studio-plugin/src/parsers/StudioSourceFileParser.ts
```diff
@@ -101,21 +101,14 @@ export default class StudioSourceFileParser {
...
-      }
-      return source;
-    };
-
     this.sourceFile.getImportDeclarations().forEach((importDeclaration) => {
       const { source } = StaticParsingHelpers.parseImport(importDeclaration);
-      if (source.endsWith(".css")) {
-        cssImports.push(getAbsolutePath(source));
+      const isRelativeCssImport =
+        source.endsWith(".css") && source.startsWith(".");
```
nmanu1, line 107: just double checking, is the reason we can't use `upath.isAbsolute` here because it considers an external package import (like `@yext/blah`) to be relative?

packages/studio-plugin/src/parsers/StudioSourceFileParser.ts
```diff
@@ -101,21 +101,14 @@ export default class StudioSourceFileParser {
...
+      if (isRelativeCssImport) {
+        cssImports.push(
```
nmanu1, line 109: we weren't previously doing anything with the page CSS imports, right? just curious if we're dropping any support by only adding local CSS imports

alextaing: We were just taking CSS imports directly from the import statement, then storing it into `cssImports`.  The only thing we did before was write these imports back to file.

I think one thing that will be ever so slightly different is if the user has non-relative imports (which we don't support yet) will not be saved to `cssImports` and not written back to the files.  However, these imports should remain, since we just modify the old page file.

This also has an effect on Layouts. Since we don't support non-relative `cssImports` in templates, we don't in Layouts either.

packages/studio-plugin/tests/sourcefiles/LayoutFile.test.ts
alextaing: done!

packages/studio-plugin/tests/__fixtures__/LayoutFile/BasicLayout.tsx
```diff
@@ -1,6 +1,5 @@
 import "./index.css";
 import ComplexBanner from "../ComponentFile/ComplexBanner";
-import "@yext/search-ui-react/index.css";
```
nmanu1, line 3: if we left this import in, would the tests break?

alextaing: I just made it so that node module imports will work, so added them back!

packages/studio-plugin/src/writers/StudioSourceFileWriter.ts
```diff
@@ -77,9 +79,13 @@ export default class StudioSourceFileWriter {
     });
     this.sourceFile.fixMissingImports();
     cssImports?.forEach((importSource) => {
-      this.sourceFile.addImportDeclaration({
-        moduleSpecifier: importSource,
-      });
+      const moduleSpecifier = upath.isAbsolute(importSource)
```
nmanu1, line 82: is this check needed because calling `getImportSpecifierWithExtension` for an npm package would result in an incorrect import path?

packages/studio-plugin/tests/sourcefiles/PageFile.getPageState.test.ts
```diff
@@ -69,13 +73,14 @@ describe("getPageState", () => {
     ]);
   });
 
-  it("correctly parses CSS imports", () => {
+  it("correctly parses CSS  imports", () => {
```
nmanu1, line 76: nit: extra space before `imports`

packages/studio-plugin/tests/sourcefiles/PageFile.getPageState.test.ts
```diff
@@ -69,13 +73,14 @@ describe("getPageState", () => {
...
     const pageFile = createPageFile("entityTemplate");
     const result = pageFile.getPageState();
 
     assertIsOk(result);
+    const expectedCssPath = getFixturePath("PageFile/index.css");
```
nmanu1, line 81: nit: this variable name is a little misleading now that we expect multiple CSS paths again

packages/studio-plugin/src/writers/StudioSourceFileWriter.ts
alextaing: yeah, it traces it back as a relative path from the import source instead of as a normal node module import

packages/studio-plugin/tests/sourcefiles/PageFile.getPageState.test.ts
alextaing: tyty

packages/studio-plugin/tests/sourcefiles/PageFile.getPageState.test.ts
alextaing: changed to `expectedIndexCssPath`!

packages/studio-plugin/tests/sourcefiles/LayoutFile.test.ts
```diff
@@ -86,8 +90,9 @@ describe("getLayoutState", () => {
 
   it("correctly parses CSS imports", () => {
     const result = getLayoutState("BasicLayout");
+    const expectedIndexCssPath = getFixturePath("LayoutFile/index.css");
```
nmanu1, line 93: is this the same as `getLayoutPath("index.css")`?

alextaing: unfortunately the `getLayoutPath` adds `.tsx` at the end

packages/studio-plugin/tests/utils/getImportSpecifier.test.ts
```diff
@@ -13,3 +16,11 @@ it("can import files in separate folders", () => {
     "../../c/d/NewFile"
   );
 });
+
+it("can preserve extensions", () => {
```
nmanu1, line 20: nit: can this test and the one above be wrapped in different `describe`s indicating which function they're testing?

alextaing: done!

nmanu1 (approved):

PR #408, opened 2023-10-11
Collapsible Sidebars

alextaing:
This PR adds collapsible sidebars to Studio.

J=SLAP-2953
TEST=auto, manual

image

image


e2e-tests/__screenshots__/custom-tailwind.spec.ts/TailwindClass-prop-editor-1.png
alextaing, line 1: Since I added the hamburger icon above this panel, it shrunk a little.

packages/studio-ui/src/components/EditorSidebar.tsx
```diff
@@ -25,16 +26,18 @@ export default function EditorSidebar(): JSX.Element {
...
-        options={Tab}
-        icons={tabIcons}
-        defaultOption={Tab.Props}
-        onSelect={setTab}
-      />
-      <Divider />
-      {renderTab(selectedTab)}
-    </div>
+    <CollapsibleSidebar side="right">
+      <div className="flex flex-col grow px-4" data-testid="EditorSidebar">
```
nmanu1, line 30: would we want this data attribute to be on `CollapsibleSidebar` now?

packages/studio-ui/src/components/CollapsibleSidebar.tsx
```diff
@@ -0,0 +1,40 @@
...
+interface CollapsibleSidebarProps {
+  side: "right" | "left";
+}
+export default function CollapsibleSidebar({
+  side,
+  children,
+}: PropsWithChildren<CollapsibleSidebarProps>): JSX.Element {
+  const [open, setOpen] = useState(true);
+  const toggleOpen = useCallback(() => {
+    setOpen(!open);
```
nmanu1, line 14: Can we use the callback for `setOpen` instead of passing in the value (e.g. something like `setOpen(isOpen => !isOpen))`? Then we would be able to remove `open` from the dependency array

packages/studio-ui/src/components/CollapsibleSidebar.tsx
```diff
@@ -0,0 +1,40 @@
...
+    "justify-end": side === "left",
+  });
+
+  const sidebarStyle = classNames("flex flex-col grow", {
+    hidden: !open,
+  });
+
+  return (
+    <div
+      className={classNames("flex flex-col", {
```
nmanu1, line 28: nit: can we do the `classNames` logic in the component body like how we compute the other styles?

packages/studio-ui/src/components/CollapsibleSidebar.tsx
alextaing: oh! forgot about this-- cool!

packages/studio-ui/src/components/CollapsibleSidebar.tsx
alextaing: yes, done!

packages/studio-ui/src/components/CollapsibleSidebar.tsx
```diff
@@ -9,32 +9,32 @@ export default function CollapsibleSidebar({
...
   return (
-    <div
-      className={classNames("flex flex-col", {
-        "w-1/4": open,
-      })}
-    >
+    <div className={sidebarStyle}>
       <div className={justifyButtonStyle}>
-        <button onClick={toggleOpen} aria-label={`Collapse ${side} sidebar`}>
+        <button onClick={toggleIsOpen} aria-label={`Collapse ${side} sidebar`}>
```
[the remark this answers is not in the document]
alextaing: yep! changed it so it says expand/collapse depending on the open state

packages/studio-ui/src/components/CollapsibleSidebar.tsx
```diff
@@ -9,32 +9,32 @@ export default function CollapsibleSidebar({
   side,
   children,
 }: PropsWithChildren<CollapsibleSidebarProps>): JSX.Element {
-  const [open, setOpen] = useState(true);
-  const toggleOpen = useCallback(() => {
-    setOpen(!open);
-  }, [open]);
+  const [isOpen, setIsOpen] = useState(true);
+  const toggleIsOpen = useCallback(() => {
```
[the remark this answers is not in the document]
alextaing: removed ☁️

packages/studio-ui/src/components/EditorSidebar.tsx
alextaing: moved it back down since we figured out that the`data-testid` does not play well with react component props.

e2e-tests/__screenshots__/add-element.spec.ts/can-add-a-Footer-component-2.png
nmanu1, line 1: the UI with the hamburger icons looks a little clunky to me. would it be a lot of work to update it? maybe to something like a clickable vertical divider between the panels

alextaing: done!

packages/studio-ui/src/components/CollapsibleSidebar.tsx
```diff
@@ -11,33 +11,43 @@ export default function CollapsibleSidebar({
 }: PropsWithChildren<CollapsibleSidebarProps>): JSX.Element {
   const [isOpen, setIsOpen] = useState(true);
   const toggleIsOpen = useCallback(() => setIsOpen((isOpen) => !isOpen), []);
-
-  const containerStyle = classNames("flex flex-col", {
+  const arrowRight =
+    (side === "right" && isOpen) || (side === "left" && !isOpen);
+  const containerStyle = classNames("flex flex-row", {
```
nmanu1, line 16: nit: I don't think we should need `flex-row` since that's the default for `flex`

packages/studio-ui/src/components/CollapsibleSidebar.tsx
```diff
@@ -0,0 +1,53 @@
...
+}: PropsWithChildren<CollapsibleSidebarProps>): JSX.Element {
+  const [isOpen, setIsOpen] = useState(true);
+  const toggleIsOpen = useCallback(() => setIsOpen((isOpen) => !isOpen), []);
+  const arrowRight =
+    (side === "right" && isOpen) || (side === "left" && !isOpen);
+  const containerStyle = classNames("flex", {
+    "w-1/4": isOpen,
+  });
+
+  const childrenStyle = classNames("flex flex-col grow", {
```
tmeyer2115, line 20: Could we move `containerStyle` and `childrenStyle` to be optional props? Something like: 

```
interface CollapsibleSidebarProps {
  getContainerStyles?: (isOpen: booelan) => string,
  getContentStyles?: (isOpen: boolean) => string
}
```

We could have defaults within the Component if these props are un-specified. I think having these props will make this Component even more generic and more re-usable for other use-cases.

alextaing: Would you be OK if we wait for a use case to come up for reusing this component?  I'm just a little worried that this might cause confusion since the `hidden` part of the `contentStyle` is integral to the component's function, and I wouldn't want this to be overwritten, I don't think.  I think if we find a use case for it, it'll give us more info on what exactly we might want to allow customization.  What are your thoughts?

tmeyer2115 (approved):

PR #409, opened 2023-10-11
Add `?inline` query param to user CSS imports

alextaing:
This PR adds a new plugin that adds the `?inline` query param to all user CSS imports.  This plugin ignores CSS imports from Studio.  The inline parameter tells Vite not to inject the CSS as a styletag into the document head.  We want this because injected CSS will affect studio styling.

J=SLAP-2912
TEST=manual


packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,49 @@
...
+        throw new Error(
+          "Styling file could not be resolved by yext-studio-style-plugin"
+        );
+      }
+    },
+  };
+}
+
+function isImportedByStudio(id: string) {
+  return !!(id.match(/.*@yext\/studio.*/) || id.match(/.*packages\/studio*/));
```
nmanu1, line 36: what are some example paths here? for the first regex, would that always be a path in node_modules? and for the second, is there some way to make sure the path is in our monorepo (e.g. check relative to the current file's path or something)?

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+function isStyleFile(id: string) {
+  return !!id.match(/.*\.s?css.*/);
```
nmanu1, line 40: We don't support `.scss` files yet, right? Is this just to make sure they're still not added as a tag?

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+function addQueryParameter(id: string, queryParameter: string) {
+  if (id.match(new RegExp(`.*?.*${queryParameter}`))) {
```
nmanu1, line 44: I think this would match something like `.../index.css?blahinline`, right? can we restrict it so the query param comes immediately after a `?` or `&` instead of any character?

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,49 @@
...
+    return id;
+  }
+  const joinCharacter = id.includes("?") ? "&" : "?";
```
nmanu1, line 47: I know `URL` has built-in functionality to make adding query params easier for urls. Is there anything similar for file paths so we don't have to do this manually?

packages/studio-plugin/src/createStudioStylingPlugin.ts
nmanu1, line 36: I think assuming `/` is the path separator could be an issue on windows. have you tested that?

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: Ahh the more I think about it, I'll just change it to `.css` as a skateboard for now.  I originally had some `scss` stuff in the earlier PRs but I removed them, just forgot to change this one! 

Once we get this working nicely with css we can include scss. It hopefully it'll just be a matter of including it in the process later.

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: an example would be something like:
`/Users/ataing/Documents/yext-marketing-creative/node_modules/@yext/studio/index.html`
It will always be in node modules!

For our monorepo check, I made it so that we derive the packages dir from our current path, and ignore all the css imports that originate from under that directory!

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: fixed!

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: No built in way I could find! We could install a library that does this, but I think that might be a little excessive?

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -32,16 +36,18 @@ export default function createStudioStylingPlugin(): PluginOption {
   };
 }
 
-function isImportedByStudio(id: string) {
-  return !!(id.match(/.*@yext\/studio.*/) || id.match(/.*packages\/studio*/));
+function isImportedByStudio(importer: string) {
+  return !!(
+    importer.match(/.*@yext[\\\\/]studio.*/) || importer.startsWith(packagesDir)
```
nmanu1, line 41: we can probably use `path.sep` to be more specific here instead of checking for either forward or back slashes. alternatively, we could run the `importer` path through `upath` to begin with and then we can be assured that the path separator with always be `/` so we wouldn't need special windows handling

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -32,16 +36,18 @@ export default function createStudioStylingPlugin(): PluginOption {
...
+  );
 }
 
 function isStyleFile(id: string) {
-  return !!id.match(/.*\.s?css.*/);
+  return !!id.match(/.*\.css.*/);
```
nmanu1, line 46: is this the same as `id.endsWith(".css")`?

alextaing: I wanted to make sure that it would work on any css file, even if we or Vite happened to add any query parameters (`e.g. ?inline` or `?import`)!

packages/studio-plugin/src/createStudioStylingPlugin.ts
nmanu1: So in the meantime, if someone imports a `.scss` file, the styling will be applied to Studio?

alextaing: yeah, that would be the case. Do you think it would be preferable if we included scss in this ignore now as well?

packages/studio-plugin/src/createStudioStylingPlugin.ts
nmanu1: can we update the first check to make sure the path is within `node_modules`?

packages/studio-plugin/src/createStudioStylingPlugin.ts
nmanu1: I think it's fine as is then. The logic isn't adding a lot of complexity

packages/studio-plugin/src/createStudioStylingPlugin.ts
nmanu1: got it. maybe the regex can check that there's optionally a `?` followed by any characters at the end? we wouldn't want it to match something like `.../index.css.tsx`. or alternatively, we could split on `?` and check that the first section ends with `.css`

packages/studio-plugin/src/createStudioStylingPlugin.ts
nmanu1: That would be my preference (since it doesn't seem to difficult to implement). We still won't support applying their styles but at least in the meantime it won't mess with our UI

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: I went with the latter!

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: will test the new unix conversion implementation on my windows laptop and will update!

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: done!

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: done!

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: went with the latter!

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,58 @@
...
+ * unless they are CSS from Studio, which is determined
+ * from the path of its importer. This prevents user styling
+ * from affecting Studio UI.
+ */
+export default function createStudioStylingPlugin(): PluginOption {
+  return {
+    name: "yext-studio-style-plugin",
+    enforce: "pre",
+    async resolveId(id, importer) {
+      if (!importer) {
```
nmanu1, line 18: when would the `importer` not be defined?

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,58 @@
...
+        return;
+      }
+      if (isImportedByStudio(importer)) {
+        return;
+      }
+      if (isStyleFile(id)) {
+        const resolvedId = (
+          await this.resolve(id, importer, { skipSelf: true })
```
nmanu1, line 26: does this resolve the imported path to an absolute file path? and does the `skipSelf` skip running the plugin's `resolveId` logic on it while resolving?

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,58 @@
...
+  const unixImporter = upath.toUnix(importer);
+  return !!(
+    unixImporter.match(/.*\/node_modules\/@yext\/studio.*/) ||
+    importer.startsWith(packagesDir)
+  );
+}
+
+function isStyleFile(id: string) {
+  const idWithoutQueryParams = id.split("?")[0];
+  return !!idWithoutQueryParams.endsWith(".css");
```
nmanu1, line 49: should this check for either `.css` or `.scss`?

packages/studio/vite.config.ts
```diff
@@ -27,7 +28,12 @@ export default defineConfig((args: ConfigEnv): UserConfig => {
         input: ["index.html", "src/store/useStudioStore.ts"],
       },
     },
-    plugins: [react(), createStudioPlugin(args), svgr() as PluginOption],
+    plugins: [
+      react(),
+      createStudioPlugin(args),
+      createStudioStylingPlugin(),
+      svgr() as PluginOption,
+    ],
```
nmanu1, line 36: out of curiosity, does the order of the plugins matter?

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: I think it's only the case for the entry point of the plug-in: https://rollupjs.org/plugin-development/#resolveid

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function addQueryParameter(id: string, queryParameter: string) {
+  if (id.match(new RegExp(`.*\\?(.*&)?${queryParameter}`))) {
```
nmanu1, line 53: what are the `\\` doing in this regex? I would have expected only one backslash to escape the question mark?

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,58 @@
...
+        );
+      }
+    },
+  };
+}
+
+function isImportedByStudio(importer: string) {
+  const unixImporter = upath.toUnix(importer);
+  return !!(
+    unixImporter.match(/.*\/node_modules\/@yext\/studio.*/) ||
```
nmanu1, line 42: looks like this would match any of our Studio packages (`studio-plugin`, `studio-ui`, or `studio`). is that what we want?

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: yep! returns an absolute path.  And yeah, that's exactly what`skipSelf` does.  It is supposed to default to `true` but I found that it doesn't.  Without that flag, it'll enter an infinite loop since it calls `resolveId` again. For more info if you're interested: https://rollupjs.org/plugin-development/#this-resolve

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: oh, yes-- thanks for the catch

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: Yeah, I was confused about that too at first. Since I'm writing the regex as a string, the backslash itself needs to be escaped.  That escaped backslash escapes the question mark.

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: Yeah, I think we want any styling from any "studio" package to be ignored by the plugin since we want normal Studio styling to be injected.  We have the `tailwind-directives` from `studio` and `react-toastify` from `studio-ui`.  I guess we don't have anything from `studio-plugin` though.

packages/studio/vite.config.ts
alextaing: I think in general, plugin order matters for Rollup and Vite, but in this case it works either before or after our `createStudioPlugin`.  We also `enforce` that this plugin is run as `pre`, which is before our `createStudioPlugin` which is unenforced (meaning that it runs after `pre` but before `post`.

packages/studio-plugin/src/createStudioStylingPlugin.ts
nmanu1: sounds good! I suppose if we ever release any other `@yext/studio-blah` repos outside of this monorepo we may want to update, but probably not important for now

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -46,7 +46,10 @@ function isImportedByStudio(importer: string) {
 
 function isStyleFile(id: string) {
   const idWithoutQueryParams = id.split("?")[0];
-  return !!idWithoutQueryParams.endsWith(".css");
+  return !!(
```
nmanu1, line 49: nit: can the double negation be removed now?

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: I realized by laptop is set up to use WSL which is just linux-- would it be OK if I run tests on the windows laptop in office monday?

packages/studio/vite.config.ts
nmanu1: If we do want this plugin to run before `createStudioPlugin`, it may be safer to switch the order in case we ever end up adding `pre` to `createStudioPlugin`

packages/studio-plugin/src/createStudioStylingPlugin.ts
nmanu1: since this is being merged into a feature branch, I'm fine with that!

packages/studio/vite.config.ts
alextaing: true, switched!

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: done!

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -2,7 +2,7 @@ import { PluginOption } from "vite";
 import upath from "upath";
 import { fileURLToPath } from "url";
 const currentDir = upath.dirname(fileURLToPath(import.meta.url));
-const packagesDir = upath.resolve(currentDir, "../..");
+const packagesDir = upath.resolve(currentDir, `..${upath.sep}..`);
```
nmanu1, line 5: pretty sure `upath.sep` will always be `/` because it resolves to unix first, so you'd actually want `path.sep`. but I think this isn't necessary (and we can go back to what you had before) since we're already running this through `upath`. you can read their docs to double check though

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -40,7 +40,7 @@ function isImportedByStudio(importer: string) {
   const unixImporter = upath.toUnix(importer);
   return !!(
     unixImporter.match(/.*\/node_modules\/@yext\/studio.*/) ||
-    importer.startsWith(packagesDir)
```
alextaing: I was concerned that on windows, the `importer` would be in windows, and the `packageDir` would be in unix.  So just to be safe, I converted them both to unix.  Do you think this is unnecessary?

I tried to find info about what `upath.resolve` does exactly but for some reason I was having trouble tracing it back...

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -40,7 +40,7 @@ function isImportedByStudio(importer: string) {
...
+    unixImporter.startsWith(packagesDir)
```
alextaing, line 43: I was concerned that on windows, the `importer` would be in windows, and the `packageDir` would be in unix.  So just to be safe, I converted them both to unix.  Do you think this is unnecessary?

I tried to find info about what `upath.resolve` does exactly but for some reason I was having trouble tracing it back...

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: ok! I also made sure that we compare this unix `packageDir` to the unix version of `importer` in the studio import check.

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -2,7 +2,7 @@ import { PluginOption } from "vite";
 import upath from "upath";
 import { fileURLToPath } from "url";
 const currentDir = upath.dirname(fileURLToPath(import.meta.url));
-const packagesDir = upath.resolve(currentDir, "../..");
+const packagesDir = upath.resolve(currentDir, `../..`);
```
nmanu1, line 5: nit: `../..` can be a regular string instead of template string now

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,61 @@
...
+      }
+    },
+  };
+}
+
+function isImportedByStudio(importer: string) {
+  const unixImporter = upath.toUnix(importer);
+  return !!(
+    unixImporter.match(/.*\/node_modules\/@yext\/studio.*/) ||
+    unixImporter.startsWith(packagesDir)
```
nmanu1, line 43: nit: if you switch the order of these two checks, you should be able to remove the double negation

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,61 @@
...
+function isStyleFile(id: string) {
+  const idWithoutQueryParams = id.split("?")[0];
+  return (
+    idWithoutQueryParams.endsWith(".css") ||
+    idWithoutQueryParams.endsWith(".scss")
+  );
+}
+
+function addQueryParameter(id: string, queryParameter: string) {
+  if (id.match(new RegExp(`.*\\?(.*&)?${queryParameter}`))) {
```
nmanu1, line 56: can we also have this regex check that this query param is at the end of the string, or there's a `&` immediately? in other words, we don't want to match `.../index.css?inlineblah`

packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -0,0 +1,61 @@
...
+    return id;
+  }
+  const joinCharacter = id.includes("?") ? "&" : "?";
```
nmanu1, line 59: could there be a case where there's a `?` or `&` at the end of the path with nothing after?

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: Out of curiosity, what is changed by the difference in ordering?  My understanding was that if the first is true, return the first, else return the second.

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: Not that I know of-- I think if that did happen that would suggest that something is wrong.

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: I think for cleanliness, it would be nice to at least negate the `match` statement so we always return a boolean.

nmanu1: ah nvm I agree

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: done!

nmanu1 (approved):

nmanu1 (approved):

PR #411, opened 2023-10-13
Support aliased type imports

nmanu1:
This PR adds support for aliased type imports like `import { CtaData as Cta } from "@yext/search-ui-react";`. This includes multiple levels of aliasing such as an aliased import that imports a type that was itself an aliased import. A number of other bugs are fixed as well:
- Fixed missing imports in general across all source files.
- A fix to make sure we check that imported Studio types in an external types file (i.e. not directly within a Component file) are properly checked when determining whether the type is imported from Studio. Previously, we only looked within the Component file for imports, so even if a Studio type was imported correctly in an external types file, we would still give an error unless an import for that type was added in the Component file (despite it being unused). The inverse was also an issue where if the type in the external file was not imported correctly, as long as the Component file imported that type from Studio, we wouldn't give an error.
- A fix to handle string literal prop types correctly (i.e. “number” is treated as a string literal with the value "number" rather than the `number` type).

One thing to note is that Studio doesn't currently support tracing through `export` statements when parsing imported types. For example, if you try to import `UserPaths` from `studio-plugin` in the test site, we will not be able to resolve the type because the type is exported from `studio-plugin` using the `export ... from ...` syntax (e.g. `export { UserPaths } from "./types";`) instead of separate import and export statements. We only trace through `import` statements for now.

J=SLAP-2683
TEST=auto, manual

See that the updated test site still resolves the type for `nestedColor` properly and the value of that prop can be changed through the UI and saved as expected.


tmeyer2115 (dismissed):

alextaing (approved):
absolute fire.

tmeyer2115 (approved):

alextaing (approved):

PR #412, opened 2023-10-16
Import CSS and Inject into Iframe

alextaing:
This PR adds user styling into the preview iframe. First, Studio dynamically imports user CSS and adds it to the head of Studio as a disabled style tag.  Then, the `IframePortal` injects the style into the preview panel depending on the page CSS and the CSS of the components in the component tree.

Note that this PR would break acceptance test screenshots since Tailwind is not currently configured in e2e-tests, so we turned off screenshots in this PR.  This will be reverted in the following PR when we add Tailwind.

J-SLAP-2912
TEST=manual


packages/studio-ui/src/utils/loadUserAssets.tsx
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Load all user styling as disabled style tags in Studio's head.
+ */
+export function loadStyling() {
+  const studioStore = useStudioStore.getState();
+  const UUIDToFileMetadata = studioStore.fileMetadatas.UUIDToFileMetadata;
+  const pages = studioStore.pages.pages;
+
+  Object.values(UUIDToFileMetadata).forEach((fileMetadata) => {
+    if (fileMetadata.kind === FileMetadataKind.Component) {
```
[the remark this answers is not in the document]
alextaing: Oh, it's just that the error components don't have their CSS imports parsed.

packages/studio-ui/src/components/IFramePortal.tsx
```diff
@@ -42,21 +42,71 @@ export default function IFramePortal(
...
+
+function clearStylingFromIframe(iframeDocument: Document) {
+  const styleElements = [...iframeDocument.head.getElementsByTagName("style")];
+  for (const el of styleElements) {
+    el.remove();
+  }
+}
+
+function injectStyleIntoIframe(
+  iframeDocument: Document | undefined,
```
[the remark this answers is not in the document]
alextaing: yes, changed!

packages/studio-ui/src/components/IFramePortal.tsx
```diff
@@ -42,21 +42,71 @@ export default function IFramePortal(
...
+    iframeDocument,
+    pageCss,
+    activePage,
+    UUIDToFileMetadata,
+  ]);
+}
+
+function clearStylingFromIframe(iframeDocument: Document) {
+  const styleElements = [...iframeDocument.head.getElementsByTagName("style")];
+  for (const el of styleElements) {
```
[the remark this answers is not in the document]
alextaing: yes, done!

packages/studio-ui/src/components/IFramePortal.tsx
alextaing, line 1: I feel like this component has become a little bloated.  What are your thoughts on extracting our styling hooks to other files?

e2e-tests/__screenshots__/undefined-menu-button.spec.ts/renders-undefined-menu-button-1.png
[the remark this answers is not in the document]
alextaing: done!

packages/studio-ui/src/utils/loadUserAssets.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  Object.values(pages).forEach((page) => {
+    importAndInjectIntoStudio(page.cssImports);
+  });
+}
+
+function importAndInjectIntoStudio(cssImports: string[]) {
+  cssImports.forEach(async (filepath) => {
+    await dynamicImportFromBrowser(filepath).then((styling) => {
+      if (document.querySelector(`[studio-style-filepath='${filepath}']`)) {
```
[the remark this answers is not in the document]
nmanu1: Not sure I follow. Is the dynamic import needed in order to check if the style tag is a duplicate?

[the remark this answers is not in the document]
alextaing: oops I misread your comment!  You're right, I was concerned about some asynchronous cases where we might be dynamically importing the same component at the same time, but since we await the dynamic import I think that those concerns are void. I'll move it up.

packages/studio-ui/src/utils/loadUserAssets.tsx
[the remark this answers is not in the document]
nmanu1: What's the reason for that? I thought we still tried to preview the component even if it's in an error state, so we'd want CSS for it if possible?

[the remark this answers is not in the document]
alextaing: Chatted offline-- decided to make an item to support Error Component CSS in a later item.  We also decided to support CSS HMR in a later item!

packages/studio-ui/src/utils/loadUserAssets.tsx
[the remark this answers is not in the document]
alextaing: As we chatted offline about, I changed the stream to a for loop to prevent race conditions

packages/studio-plugin/tests/sourcefiles/ComponentFile.test.ts
```diff
@@ -177,7 +177,7 @@ describe("getComponentMetadata", () => {
 
   it("correctly parses a string literal type", () => {
     const pathToComponent = getComponentPath("StringLiteralBanner");
-    const componentFile = new ComponentFile(pathToComponent, project);
+    const componentFile = new ComponentFile(pathToComponent, project, {});
```
nmanu1, line 180: Were these changes from a previous PR?

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -0,0 +1,62 @@
...
+import useStudioStore from "../store/useStudioStore";
+
+export default function useInjectUserStyles(
+  iframeDocument: Document | undefined
+) {
+  const [
+    componentTree,
+    getComponentMetadata,
+    activePage,
+    UUIDToFileMetadata,
```
nmanu1, line 11: where are `activePage` and `UUIDToFileMetadata` used?

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -0,0 +1,62 @@
...
+    pageCss,
+  ] = useStudioStore((store) => [
+    store.actions.getComponentTree(),
+    store.actions.getComponentMetadata,
+    store.pages.activePageName,
+    store.fileMetadatas.UUIDToFileMetadata,
+    store.pages.getActivePageState()?.cssImports,
```
nmanu1, line 18: out of curiosity, if the active page changes, do we see multiple extra re-renders because we're subscribing to different parts of the active page state separately? or would it be the same as if we subscribed to the whole active page state at once?

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -0,0 +1,62 @@
...
+    pageCss,
+    activePage,
+    UUIDToFileMetadata,
+  ]);
+}
+
+function clearStylingFromIframe(iframeDocument: Document) {
+  const styleElements = [...iframeDocument.head.getElementsByTagName("style")];
+  styleElements.forEach((element) => {
+    element.remove();
```
nmanu1, line 50: if there was studio styling for a component within the iframe, could this be a problem?

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -0,0 +1,62 @@
...
+  });
+}
+
+function injectStyleIntoIframe(iframeDocument: Document, filepath: string) {
+  const originalStyletag = document.querySelector(
+    `[studio-style-filepath='${filepath}']`
+  );
+  if (originalStyletag) {
```
nmanu1: would we want to log some kind of warning if the styletag isn't found? what are the scenarios in which that could happen?

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -0,0 +1,63 @@
...
+function clearStylingFromIframe(iframeDocument: Document) {
+  const styleElements = [...iframeDocument.head.getElementsByTagName("style")];
+  styleElements.forEach((element) => {
+    element.remove();
+  });
+}
+
+function injectStyleIntoIframe(iframeDocument: Document, filepath: string) {
+  const styletagIdAttribute = `[studio-style-filepath='${filepath}']`;
+  const originalStyletag = document.querySelector(styletagIdAttribute);
```
nmanu1, line 56: nit: can this variable name be changed to indicate that this is the style tag from the parent document? I was a little confused at first

packages/studio-ui/src/AppWithLazyLoading.tsx
```diff
@@ -18,7 +18,7 @@ export default function AppWithLazyLoading() {
 
   useEffect(() => {
     loadComponents();
```
nmanu1, line 20: should we also add `void` before `loadComponents`? that must also be doing async work to load in the components

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
alextaing: They aren't used explicitly, but I wanted to call the useEffect hook to update the style when we change the active page or FileMetadatas are updated!  But I think you have a point about the active page one.  That's not necessary since we're already watching if the page CSS is changing directly.  The filemetata data check we will still need in case users change their CSS for HMR.

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
alextaing: I removed the activePageName, so we shouldn't run into issues now!

nmanu1: oh I meant that we get the active page's component tree and css imports separately

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
nmanu1: Does `getComponentMetadata` change if `UUIDToFileMetadata` changes in the store? And also, I thought we weren't handling HMR for CSS?

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
alextaing: There should never be a studio styling style tag in the iframe!  We only move styletags into the iframe that are imported by the user.  If the user and Studio happen to be using the same CSS, there would be two versions of the styletag in Studio's (parent) head (one from vite that we ignore using the `?studioStyling` query param, and one with our custom attribute, `studio-user-custom-style`. The only styletags that we care about for the iframe are the ones with `studio-user-custom-style`.

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
alextaing: ahh that's a better word for it.  I was struggling to find the right word.  Yes, I'll change that!

packages/studio-plugin/tests/sourcefiles/ComponentFile.test.ts
alextaing: I think when you added tests for the alias types, you also added new tests that used `ComponentFile`. I've been doing my best to keep the feature branch up to date with main, so I just updated the new tests to work with these new changes.  This additional parameter is from where we pass our dependency tree (for component file CSS).

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
alextaing: Unfortunately, I found that `getComponentMetadata` doesn't change when `UUIDToFileMetadata` changes.

For HMR, we currently have 2 potential use cases: 
1) the user updates a component file by adding or removing CSS imports.  We can currently support HMR for this, unless the file is a completely new CSS file.  That is just limited by the fact that we haven't added the styletag to Studio's parent head since we only do that once on the lazy load.
2) the user changes a CSS file, which is the thing we can't support and Tom says we don't care about at the moment haha

Do you think this would be confusing and better to not support at all, or is it the case that better some than none?

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
alextaing: The only time where that would be the case is if the user adds a completely new CSS import to a page or component.  In this case, the filemetadata will be updated through component HMR, but since we only add styletags to Studio's parent head at the beginning of Studio load, a corresponding styletag for the new CSS file won't be found. I can add a warning 👍

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
alextaing: Ah, I see. I tested it out and based on what I could tell, the iframe renders the same amount (twice) either way.  I changed it to just import the active page state once, since that seems a little cleaner!

packages/studio-ui/src/AppWithLazyLoading.tsx
alextaing: I was also considering that.  The way that Oliver implemented this was he returned a list of void promises from the function, but I'm not entirely sure what the reasoning behind it was.

I think that the way the functions are implemented, though, are slightly different where we don't necessarily need the void on `loadComponents`.  For our `loadStyling` we needed to write the function very imperatively, so that we can properly await each dynamic import before beginning the next. This meant that the whole `loadStyling` function was async.  But, for `loadComponents` each dynamic import is independent of another, so Oliver called each dynamic import in a `map`, which as a whole didn't need to be awaited. So, `loadComponents` is not an "async" function itself.

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
nmanu1: So even if a Studio component is only used within `IFramePortal`, Vite would still add the styletags for its CSS (because it would have `?studioStyling`) to the parent document's head rather than the iframe's?

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
nmanu1: If the number of renders is the same, I'd actually prefer keeping them as separate selectors. Because, while it is a little cleaner to group them, we'd get extra re-renders if some part of the PagesJS state for the active page was changed even though that doesn't affect the styling

alextaing: ahhh that makes sense. I'll change it

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
nmanu1: I left another comment that includes a suggestion for how we can get around watching `UUIDToFileMetadata` and adding it to the dependency array. It would be nice to have HMR support if it "comes out of the box" almost, even if it is limited. My only hesitation is that I prefer not to rely on the presence of an unused dependency in the dep array to trigger updates because it would be easy down the road to forget why it was included and remove it since it's not actually used in the callback. Another alternative could be to just add a comment explaining why it's included in the dep array

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -1,26 +1,23 @@
 import { useEffect } from "react";
 import useStudioStore from "../store/useStudioStore";
 
+export const UserCustomStyletagAttribute = "studio-user-custom-style";
```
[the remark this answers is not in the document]
alextaing: move and renamed to all caps 👍

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -52,12 +47,19 @@ function clearStylingFromIframe(iframeDocument: Document) {
...
 function injectStyleIntoIframe(iframeDocument: Document, filepath: string) {
-  const styletagIdAttribute = `[studio-style-filepath='${filepath}']`;
-  const originalStyletag = document.querySelector(styletagIdAttribute);
+  const styletagIdAttribute = `[${UserCustomStyletagAttribute}='${filepath}']`;
+  const parentDocumentStyletag = document.querySelector(styletagIdAttribute);
   const oldIframeStyletag = iframeDocument.querySelector(styletagIdAttribute);
+  if (!parentDocumentStyletag) {
+    console.warn(
+      `${filepath} was not able to be loaded into the Studio Preview. ` +
+        "This is likely because it is a newly added CSS file. Refresh Studio to update."
```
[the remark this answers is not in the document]
alextaing: reworded to this!

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -0,0 +1,66 @@
...
+import useStudioStore from "../store/useStudioStore";
+
+export const UserCustomStyletagAttribute = "studio-user-custom-style";
+
+export default function useInjectUserStyles(
+  iframeDocument: Document | undefined
+) {
+  const [componentTree, getComponentMetadata, UUIDToFileMetadata, pageCss] =
+    useStudioStore((store) => [
+      store.actions.getComponentTree(),
```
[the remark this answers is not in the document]
alextaing: Chatted offline, we computed the list of CSS files from the component tree and only subscribe the `useEffect` to the CSS files instead of the component tree.

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
alextaing: Chatted offline, see[ this comment](https://github.com/yext/studio/pull/412#discussion_r1366103331), where we remove the need for subscribing to `UUIDToFileMetadata`

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -216,12 +227,13 @@ export default class ParsingOrchestrator {
...
-  private updateFilepathToDependencyTree(absPath: string) {
-    this.filepathToDependencyTree[absPath] = dependencyTree({
+  private updateDependencyTrees(absPath: string) {
+    const newDepTree = dependencyTree({
       filename: absPath,
       directory: upath.dirname(absPath),
-      visited: this.filepathToDependencyTree,
+      visited: this.dependencyTrees,
     });
+    this.dependencyTrees = Object.assign(this.dependencyTrees, newDepTree);
```
nmanu1, line 236: could we do a spread here instead? I personally find that a little more readable

packages/studio-plugin/src/parsers/StudioSourceFileParser.ts
```diff
@@ -123,19 +124,29 @@ export default class StudioSourceFileParser {
   parseCssImports(): string[] {
     const cssImports: string[] = [];
 
-    const resolveRelativeFilepath = (filepath: string) => {
-      if (filepath.startsWith(".")) {
-        return upath.resolve(
-          upath.dirname(this.sourceFile.getFilePath()),
-          filepath
+    const getImportAbsoluteFilepath = (importPath: string) => {
```
nmanu1, line 127: what do you think of `getAbsoluteImportFilepath` instead?

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
```diff
@@ -62,8 +63,9 @@ export default class ComponentFile {
 function getCssFilesFromDependencyTree(dependencyTree: Tree): string[] {
   const cssFiles = Object.entries(dependencyTree).reduce(
     (cssFiles, [filename, subDependencyTree]) => {
-      if (filename.includes(".css")) {
-        cssFiles.add(filename);
+      const unixFilepath = upath.toUnix(filename);
+      if (unixFilepath.includes(".css")) {
```
nmanu1, line 67: should this check that it ends with `.css` rather than includes it?

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -113,7 +113,12 @@ export default class ParsingOrchestrator {
     }
 
     if (filepath.startsWith(this.paths.components)) {
-      delete this.filepathToDependencyTree[filepath];
+      const componentDepTreeRoot = Object.keys(this.dependencyTrees).find(
+        (path) => upath.toUnix(path) === filepath
```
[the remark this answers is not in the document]
alextaing: done!

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -9,38 +11,45 @@ import { USER_CUSTOM_STYLE_ATTRIBUTE } from "../utils/loadUserAssets";
 export default function useInjectUserStyles(
   iframeDocument: Document | undefined
 ) {
-  const [componentTree, getComponentMetadata, UUIDToFileMetadata, pageCss] =
-    useStudioStore((store) => [
-      store.actions.getComponentTree(),
-      store.actions.getComponentMetadata,
-      store.fileMetadatas.UUIDToFileMetadata,
+  const [componentCss, pageCss] = useStudioStore(
```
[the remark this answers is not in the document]
alextaing: done!

packages/studio-plugin/src/writers/StudioSourceFileWriter.ts
```diff
@@ -79,12 +78,12 @@ export default class StudioSourceFileWriter {
     });
     this.sourceFile.fixMissingImports();
     cssImports?.forEach((importSource) => {
-      const moduleSpecifier = upath.isAbsolute(importSource)
-        ? getImportSpecifierWithExtension(
+      const moduleSpecifier = importSource.includes("/node_modules/")
+        ? importSource.split("/node_modules/")[1]
```
nmanu1, line 82: instead of checking if it includes `/node_modules/` and then splitting on it, can we always split on it and if the length of the resulting array is more than 1, return the second item?

packages/studio-plugin/tests/__fixtures__/LayoutFile/BasicLayout.tsx
```diff
@@ -1,6 +1,5 @@
 import "./index.css";
 import ComplexBanner from "../ComponentFile/ComplexBanner";
-import "@yext/search-ui-react/index.css";
```
nmanu1, line 3: can we try testing with `@yext/search-ui-react/bundle.css` if you need it to resolve to a real path?

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
alextaing: done!

packages/studio-plugin/src/parsers/StudioSourceFileParser.ts
alextaing: I like it!

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
alextaing: yes!

packages/studio-plugin/src/writers/StudioSourceFileWriter.ts
alextaing: Do you mean something like
```
     const splitOnNodeModules = importSource.split("/node_modules/");
      const moduleSpecifier = splitOnNodeModules.length > 1
        ? splitOnNodeModules[1]
        : getImportSpecifierWithExtension(
            this.sourceFile.getFilePath(),
            importSource
          );
```

If so, I think I prefer it's current state. Do you prefer this one?

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -36,12 +36,13 @@ export function createTsMorphProject(tsConfigFilePath: string) {
  */
 export default class ParsingOrchestrator {
   private filepathToFileMetadata: Record<string, FileMetadata> = {};
-  private dependencyTrees: Record<string, Tree> = {};
   private pageNameToPageFile: Record<string, PageFile> = {};
   private siteSettingsFile?: SiteSettingsFile;
   private studioData?: StudioData;
   private paths: UserPaths;
   private layoutOrchestrator: LayoutOrchestrator;
+  /** This tree contains a root for each Component File. */
```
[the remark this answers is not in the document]
alextaing: improved the doc!

packages/studio-plugin/tests/__fixtures__/LayoutFile/BasicLayout.tsx
alextaing: Discussed offline, will address this in another PR!

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
```diff
@@ -62,8 +63,9 @@ export default class ComponentFile {
 function getCssFilesFromDependencyTree(dependencyTree: Tree): string[] {
   const cssFiles = Object.entries(dependencyTree).reduce(
     (cssFiles, [filename, subDependencyTree]) => {
-      if (filename.includes(".css")) {
-        cssFiles.add(filename);
+      const unixFilepath = upath.toUnix(filename);
```
nmanu1, line 66: can this conversion go inside the if-statement below since it's only needed if the file is a CSS file?

nmanu1, line 66: to clarify, is the `filename` variable here actually an absolute file path?

packages/studio-plugin/src/writers/StudioSourceFileWriter.ts
nmanu1: Yes, that's what I was imagining. I slightly prefer it because it feels a little redundant for us to first check for `node_modules` occurring in the string and then doing the split, which also has to check for where `node_modules` occurs in the string. But, if you'd prefer keeping the existing logic, then can we pull out `/node_modules/` into a constant since it's used more than once?

packages/studio-plugin/tests/__fixtures__/PageFile/brokenCssImport.tsx
```diff
@@ -0,0 +1,22 @@
...
+  },
+};
+
+export const getPath: GetPath<TemplateProps> = ({
+  document,
+}: TemplateProps) => {
+  return document.slug;
+};
+
+export default function BrokenCss({ document }: TemplateProps) {
```
nmanu1, line 20: nit: can this be called unsupported CSS instead of broken? it's not that there's something wrong with the CSS. we just don't support it right now

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -0,0 +1,74 @@
...
+ * Studio Preview iframe's document head.  It only injects the relevant
+ * styling of the current active page and component tree.  Styling is cleared
+ * from the iframe before each injection.
+ */
+export default function useInjectUserStyles(
+  iframeDocument: Document | undefined
+) {
+  const userCssImports = useStudioStore(
+    (store) => getUserCssImports(store),
+    isEqual
```
nmanu1, line 18: could you make a code maintenance item to investigate using `isEqual` as the default comparator for `useStudioStore`? I think that could be useful for reducing re-renders

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const cssImports = getComponentMetadata(component)?.cssImports;
+    cssImports?.forEach((cssFilepath) => {
+      componentTreeCss.add(cssFilepath);
+    });
+  });
+  return [...componentTreeCss, ...pageCss];
+}
+
+function clearStylingFromIframe(iframeDocument: Document) {
+  const styleElements = [...iframeDocument.head.getElementsByTagName("style")];
```
nmanu1, line 51: what's the reason for the spread here?

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -0,0 +1,74 @@
...
+  styleElements.forEach((element) => {
+    element.remove();
+  });
+}
+
+function injectStyleIntoIframe(iframeDocument: Document, filepath: string) {
+  const styletagIdAttribute = `[${USER_CUSTOM_STYLE_ATTRIBUTE}='${filepath}']`;
+  const parentDocumentStyletag = document.querySelector(styletagIdAttribute);
+  const oldIframeStyletag = iframeDocument.querySelector(styletagIdAttribute);
+  if (!parentDocumentStyletag) {
```
nmanu1, line 61: can we do a check for `oldIframeStyletag` before this check? that way, if there's already a tag present in the iframe, we can return early since we don't care if there's a corresponding one on the parent document

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const styleElements = [...iframeDocument.head.getElementsByTagName("style")];
+  styleElements.forEach((element) => {
+    element.remove();
+  });
+}
+
+function injectStyleIntoIframe(iframeDocument: Document, filepath: string) {
+  const styletagIdAttribute = `[${USER_CUSTOM_STYLE_ATTRIBUTE}='${filepath}']`;
+  const parentDocumentStyletag = document.querySelector(styletagIdAttribute);
+  const oldIframeStyletag = iframeDocument.querySelector(styletagIdAttribute);
```
nmanu1, line 60: nit: can this be called something like `existingIframeStyletag`? with the current naming, I expected that an "old" tag would be out-of-date and we'd try to replace it with a new one

packages/studio-ui/src/AppWithLazyLoading.tsx
```diff
@@ -18,6 +18,7 @@ export default function AppWithLazyLoading() {
...
+    void loadStyling();
```
nmanu1, line 21: do we need a `stylingLoaded` check like we have `componentsLoaded`? otherwise I imagine the Studio app could be rendered before the styling finishes loading?

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
alextaing: done!

packages/studio-plugin/src/writers/StudioSourceFileWriter.ts
alextaing: OK, extracted to a variable--  is the convention of all caps only for exported constants?

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
alextaing: yes, that's correct-- I changed the name to `absFilepath` to reflect that it's a path and not just a name.

packages/studio-ui/src/utils/loadUserAssets.tsx
```diff
@@ -0,0 +1,74 @@
...
+    if (fileMetadata.kind === FileMetadataKind.Component) {
+      await importAndInjectIntoStudio(fileMetadata.cssImports);
+    }
+  }
+
+  for (const page of Object.values(pages)) {
+    await importAndInjectIntoStudio(page.cssImports);
+  }
+
+  for (const layout of Object.values(layouts)) {
```
nmanu1, line 54: is there a way to consolidate some of these for loops? they're very similar

packages/studio-plugin/src/writers/StudioSourceFileWriter.ts
nmanu1: I think we generally like the names of all constants like this to be in all caps

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
alextaing: done! https://yexttest.atlassian.net/browse/SLAP-2974

packages/studio-ui/src/hooks/useInjectUserStyles.tsx
alextaing: `iframeDocument.head.getElementsByTagName("style")` returns a "live" array, so if we try to delete from it and iterate at the same time as we normally do, it'll miss entries.  I just spread to turn it into a standard array, then iterate through it.  Some alternatives are iterating backwards, or just popping from the list until it's empty.  I figured this was the cleanest way!

packages/studio-ui/src/utils/loadUserAssets.tsx
alextaing: consolidated 👍

packages/studio-ui/src/utils/loadUserAssets.tsx
```diff
@@ -72,3 +66,11 @@ async function importAndInjectIntoStudio(cssImports: string[]) {
     });
   }
 }
+
+function getCssImportsFromUserFiles(
+  cssImporters: (ComponentMetadata | PageState | LayoutState)[]
+) {
+  return Object.values(cssImporters).reduce((cssImports, importer) => {
```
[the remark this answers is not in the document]
alextaing: done!

packages/studio-ui/src/utils/loadUserAssets.tsx
```diff
@@ -0,0 +1,76 @@
...
+      document.head.appendChild(styleEl);
+      styleEl.disabled = true;
+    });
+  }
+}
+
+function getCssImportsFromUserFiles(
+  cssImporters: (ComponentMetadata | PageState | LayoutState)[]
+) {
+  return Object.values(cssImporters).reduce((cssImports, importer) => {
```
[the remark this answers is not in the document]
alextaing: you are very right, missed that!

packages/studio-ui/src/utils/loadUserAssets.tsx
```diff
@@ -37,32 +42,21 @@ export function loadComponents(): Promise<void>[] {
...
-      document.querySelector(`[${USER_CUSTOM_STYLE_ATTRIBUTE}='${filepath}']`)
-    ) {
-      return;
-    }
+  const userCssFilepaths = getCssImportsFromUserFiles([
+    ...pages,
+    ...layouts,
+    ...componentMetadatas,
+  ]);
+  for (const filepath of userCssFilepaths) {
```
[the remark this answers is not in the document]
alextaing: I was thinking about this, too.  I guess since we're not addressing the wait for `stylingLoaded` in this PR, we don't need to await on it, and we can just use a stream function. Changed!

packages/studio-ui/src/utils/loadUserAssets.tsx
```diff
@@ -37,32 +42,21 @@ export function loadComponents(): Promise<void>[] {
...
-    await importAndInjectIntoStudio(layout.cssImports);
-  }
-}
+  const pages = Object.values(studioStore.pages.pages);
+  const layouts = Object.values(studioStore.layouts.layoutNameToLayoutState);
+  const fileMetadatas = Object.values(
+    studioStore.fileMetadatas.UUIDToFileMetadata
+  );
+  const componentMetadatas = fileMetadatas.filter(
+    (fileMetadata) => fileMetadata.kind === FileMetadataKind.Component
```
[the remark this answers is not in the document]
alextaing: done!

packages/studio-ui/src/AppWithLazyLoading.tsx
alextaing: Will address in a future PR!

nmanu1 (approved):

PR #415, opened 2023-10-19
Support the Required utility type for a prop

nmanu1:
This PR adds support for the `Required` TypeScript utility type when defining the type of a prop in a component's prop interface. As mentioned in the previous PR, utility types are still not allowed when defining the overall type of the component's prop interface (i.e. `props: Required<MyProps>`).

J=SLAP-2969
TEST=auto, manual

In the test site, see that using `Required<ObjectProp>` instead of `ObjectProp` for the type of `obj` in `BannerData` correctly results in the prop editors for `nestedString`, `nestedBool`, and `nestedObj` no longer displaying the undefined menu button icon in the UI, while all other prop editors were unchanged.

### DIFF
--- a/packages/studio-plugin/src/parsers/helpers/UtilityTypeParsingHelper.ts
+++ b/packages/studio-plugin/src/parsers/helpers/UtilityTypeParsingHelper.ts
@@ -24,6 +24,8 @@ export default class UtilityTypeParsingHelper {
           return this.handleOmit;
         case "Pick":
           return this.handlePick;
+        case "Required":
+          return this.handleRequired;
       }
     })();
     return utilityTypeHandler?.(typeArgs, parseTypeNode);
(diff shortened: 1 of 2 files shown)


packages/studio-plugin/src/parsers/helpers/UtilityTypeParsingHelper.ts
```diff
@@ -99,4 +101,39 @@ export default class UtilityTypeParsingHelper {
...
+    if (typeArgs.length !== 1) {
+      throw new Error(
+        `One type param expected for Required utility type. Found ${typeArgs.length}.`
+      );
+    }
+
+    const originalType = parseTypeNode(typeArgs[0]);
+
+    if (originalType.kind !== ParsedTypeKind.Object) {
+      return originalType;
```
tmeyer2115, line 118: Is it valid TypeScript to have Required applied to a Type that's not an Object (for example a `string` or `string` union)? If that isn't valid TypeScript, I'd say we shouldn't handle gracefully here.

nmanu1: Something like `Required<string>` is valid TypeScript. Based on my testing, it doesn't give any errors and it matches the original type. Out of the utility types I've implemented, the only one with unpredictable behavior for non-Object types was `Pick`, which is why I opted to throw an error there

PR #417, opened 2023-10-19
Add Undo Keyboard Shortcut 

alextaing:
This PR adds the undo keyboard shortcut for Windows (`ctrl+z`) and Mac (`cmd+z`) depending on the user's OS.

Note that we `preventDefault` behavior for this shortcut.  In the context of Studio in Storm, this means that when the user is focused on the Studio iframe, the default behavior will be overridden by Studio `undo`.  But, when the user clicks out of the Studio iframe, default behavior will resume.

J=SLAP-2954
TEST=manual, auto, SIS

### DIFF
--- a/packages/studio-ui/tests/components/UndoRedo.test.tsx
+++ b/packages/studio-ui/tests/components/UndoRedo.test.tsx
@@ -56,7 +56,7 @@ describe("Undo/redo", () => {
     expect(useStudioStore.getState().pages.activeComponentUUID).toBeUndefined();
   });
 
-  it("only undoes last state update using control + z if not in OS X", async () => {
+  it("undoes last state update using control + z if OS is not OS X", async () => {
     platform.os.family = "Windows";
     render(<UndoRedo />);
     expect(useStudioStore.getState().pages.activeComponentUUID).toBe(
@@ -70,7 +70,7 @@ describe("Undo/redo", () => {
     expect(useStudioStore.getState().pages.activeComponentUUID).toBeUndefined();
   });
 
-  it("only undoes last state update using command + z if in OS X", async () => {
+  it("undoes last state update using command + z if OS is OS X", async () => {
     platform.os.family = "OS X";
     render(<UndoRedo />);
     expect(useStudioStore.getState().pages.activeComponentUUID).toBe(


packages/studio-ui/tests/components/UndoRedo.test.tsx
```diff
@@ -46,6 +46,24 @@ describe("Undo/redo", () => {
...
+    expect(useStudioStore.getState().pages.activeComponentUUID).toBeUndefined();
+  });
+
+  it("undoes last state update using command + z", async () => {
+    render(<UndoRedo />);
+    expect(useStudioStore.getState().pages.activeComponentUUID).toBe(
+      "searchbar-uuid"
+    );
+    await userEvent.keyboard('{Meta>}z{/Meta}')
+    expect(useStudioStore.getState().pages.activeComponentUUID).toBeUndefined();
```
alextaing, line 64: Note that since the reference to `undo` does not change, this `useEffect` will only be run on first render.  Furthermore (fun fact), if you call `addEventListener` with a named function, it will not be added more than once.  So, if there was a case that the `useEffect` hook would be called twice, it would not add a duplicate listener.

packages/studio-ui/src/components/UndoRedo.tsx
```diff
@@ -22,6 +22,20 @@ export default function UndoRedo(): JSX.Element {
...
+      event.preventDefault();
+      if ((event.ctrlKey || event.metaKey) && event.key === "z") {
+        undo();
+      }
+    },
+    [undo]
+  );
+
+  useEffect(() => {
+    document.addEventListener("keydown", handleUndoKeydown);
```
alextaing, line 36: Note that since the reference to `undo` does not change, this `useEffect` will only be run on first render.  Furthermore (fun fact), if you call `addEventListener` with a named function, it will not be added more than once.  So, if there was a case that the `useEffect` hook would be called twice, it would not add a duplicate listener.

packages/studio-ui/src/components/UndoRedo.tsx
```diff
@@ -22,6 +22,20 @@ export default function UndoRedo(): JSX.Element {
     redo();
   }, [redo]);
 
+  const handleUndoKeydown = useCallback(
+    (event: KeyboardEvent) => {
+      if ((event.ctrlKey || event.metaKey) && event.key === "z") {
```
tmeyer2115, line 27: Talked with Alex offline. He's going to investigate the `Platform.js` library. This library supposedly can tell us which OS a browser is being used with. It's a very popular, stable library. If it works, we can be more precise here and not set two key bindings on both MacOS and Windows.

packages/studio-ui/src/components/UndoRedo.tsx
```diff
@@ -22,6 +22,20 @@ export default function UndoRedo(): JSX.Element {
...
+        event.preventDefault();
+        undo();
+      }
+    },
+    [undo]
+  );
+
+  useEffect(() => {
+    document.addEventListener("keydown", handleUndoKeydown);
```
nmanu1, line 36: can we return here so the `useEffect` cleans up the listener on unmount?

packages/studio-ui/src/components/UndoRedo.tsx
alextaing: Works locally and in Storm! Very cool-- now, the shortcut uses command if the user OS is OS X, and defaults to control if not.

packages/studio-ui/src/components/UndoRedo.tsx
alextaing: Yes! I had it at one point, not sure why I removed it-- its back now, though!

packages/studio-ui/.size-limit.cjs
```diff
@@ -1,7 +1,7 @@
 module.exports = [
   {
     path: "lib/**/*.js",
-    limit: "850 kB",
+    limit: "870 kB",
```
nmanu1, line 4: is this bump because of the new `platform` dep?

packages/studio-ui/tests/components/UndoRedo.test.tsx
```diff
@@ -46,6 +56,26 @@ describe("Undo/redo", () => {
     expect(useStudioStore.getState().pages.activeComponentUUID).toBeUndefined();
   });
 
+  it("undoes last state update using control + z if OS is not OS X", async () => {
```
nmanu1, line 59: can we also add (or update tests) tests to check that control-z won't work on OS and cmd-z won't work on Windows?

nmanu1 (commented):
can the PR description be updated?

packages/studio-ui/.size-limit.cjs
tmeyer2115: Yep!

alextaing: that's right! Before the size was around 846 kB, but with the new dep it's around 868 kB.

nmanu1: Got it! It's not a big bump, but out of curiosity, is there an import we can use to narrow down what part of `platform` we import? (e.g. like `lodash/isEqual` instead of `lodash`)

packages/studio-ui/tests/components/UndoRedo.test.tsx
alextaing: done!

packages/studio-ui/.size-limit.cjs
alextaing: not that I could find, unfortunately

tmeyer2115 (approved):

PR #418, opened 2023-10-23
Display User CSS and SCSS in Preview Panel

alextaing:
At long last, this PR allows users to view their CSS and SCSS styling in the Preview Panel. This base of this feature is complete, however, it does have a known edge case which we have a maintenance item to address.  When using third party styling imports, Studio will not resolve aliased file paths to style files.  See [this thread](https://github.com/yext/studio/pull/412#discussion_r1368648153).

J=SLAP-2912


nmanu1 (approved):
🚀

PR #419, opened 2023-10-23
Add SCSS support

alextaing:
This PR adds SCSS support by showing SCSS styling in Studio's Preview Panel.

J=SLAP-2912
TEST=manual, auto


packages/studio-plugin/src/createStudioStylingPlugin.ts
```diff
@@ -1,8 +1,9 @@
 import { PluginOption } from "vite";
+import isStyleFile from "./utils/isStyleFile";
 
 /**
- * Adds the `?inline` query parameter to all CSS imports
- * unless they are CSS imported by Studio. Studio CSS files are
+ * Adds the `?inline` query parameter to all CSS and SCSS imports
+ * unless they are imported by Studio. Studio CSS files are
```
nmanu1, line 6: nit: since you specifically call out CSS and SCSS imports separately above, the mention of just CSS here makes me think the `studioStyling` query param would only be on CSS files

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
```diff
@@ -49,29 +50,29 @@ export default class ComponentFile {
     };
 
     const filepath = this.studioSourceFileParser.getFilepath();
-    const cssImports = getCssFilesFromDependencyTree(this.dependencyTree);
+    const styleImports = getStyleFilesFromDependencyTree(this.dependencyTree);
```
nmanu1, line 53: can we update tests to check that we get SCSS files too now?

packages/studio-plugin/src/createStudioStylingPlugin.ts
alextaing: changed to "Studio styling"

packages/studio-plugin/src/sourcefiles/ComponentFile.ts
alextaing: added tests!

tmeyer2115 (approved):

PR #420, opened 2023-10-24
Fix User Styling Race Condition

alextaing:
This PR fixes the race condition where the `useInjectUserStyles` hook attempts to inject user styles from Studio's parent head before these styles have loaded.  This issue was resolved by following a similar pattern we use for component loading.  We keep track of all the styling we have imported in a set called `loadedStyling`.  As this styling array is watched by the `useEffect` in `useInjectUserStyles` so that the iframe preview panel is always in sync with styling added to the parent head.  

One thing that was considered was that when the styles are lazily dynamically imported on Studio's first load, this `useEffect` will be called multiple times.  However, since we have the loading overlay, it won't affect the user experience much.  Since `loadedStyling` shouldn't change in any other case (we don't support new styling file HMR yet), it should not trigger excessive calls past the initial load state.

J=SLAP-2976
TEST=manual (local and Storm)

### DIFF
--- a/packages/studio-ui/src/AppWithLazyLoading.tsx
+++ b/packages/studio-ui/src/AppWithLazyLoading.tsx
@@ -26,7 +26,7 @@ export default function AppWithLazyLoading() {
 
   return (
     <LoadingOverlay
-      loading={!componentsLoaded || !resourcesLoaded}
+      loading={!(componentsLoaded && stylesLoaded && appLoaded)}
       overlay={
         <>
           {renderComponentLoadingProgress(
--- a/packages/studio-ui/src/AppWithLazyLoading.tsx
+++ b/packages/studio-ui/src/AppWithLazyLoading.tsx
@@ -26,7 +26,7 @@ export default function AppWithLazyLoading() {
 
   return (
     <LoadingOverlay
-      loading={!(componentsLoaded && stylesLoaded && appLoaded)}
+      loading={!(componentsLoaded && resourcesLoaded )}
       overlay={
         <>
           {renderComponentLoadingProgress(
--- a/packages/studio-ui/src/AppWithLazyLoading.tsx
+++ b/packages/studio-ui/src/AppWithLazyLoading.tsx
@@ -26,7 +26,7 @@ export default function AppWithLazyLoading() {
 
   return (
     <LoadingOverlay
-      loading={!(componentsLoaded && resourcesLoaded )}
+      loading={!(componentsLoaded && resourcesLoaded)}
       overlay={
         <>
           {renderComponentLoadingProgress(
--- a/packages/studio-ui/src/AppWithLazyLoading.tsx
+++ b/packages/studio-ui/src/AppWithLazyLoading.tsx
@@ -15,16 +15,18 @@ export default function AppWithLazyLoading() {
   ]);
   const componentsLoaded = loadedCount === totalCount;
   const [appLoaded, setAppLoaded] = useState(false);
+  const [stylesLoaded, setStylesLoaded] = useState(false);
+  const resourcesLoaded = appLoaded && stylesLoaded;
 
   useEffect(() => {
     loadComponents();
-    loadStyling();
+    void Promise.all(loadStyling()).then(() => setStylesLoaded(true));
     void AppPromise.then(() => setAppLoaded(true));
   }, []);
 
   return (
     <LoadingOverlay
-      loading={!componentsLoaded || !appLoaded}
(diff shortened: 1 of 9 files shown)


packages/studio-ui/src/AppWithLazyLoading.tsx
```diff
@@ -15,24 +15,26 @@ export default function AppWithLazyLoading() {
   ]);
   const componentsLoaded = loadedCount === totalCount;
   const [appLoaded, setAppLoaded] = useState(false);
+  const [stylesLoaded, setStylesLoaded] = useState(false);
+  const resourcesLoaded = appLoaded && stylesLoaded;
 
   useEffect(() => {
     loadComponents();
-    loadStyling();
+    void Promise.all(loadStyling()).then(() => setStylesLoaded(true));
```
tmeyer2115, line 23: If a Component or Styling import fails, the loading screen will be displayed indefinitely. I don't think that's something we need to fix as part of this PR. We need to get some guidance from Product & UX on how to handle this case. The hanging of the loading screen when a Component import failed existed since before this PR.

packages/studio-ui/src/store/models/slices/LoadingProgressSlice.ts
```diff
@@ -0,0 +1,4 @@
+export default interface LoadingProgressSlice {
+  loadedStyles: Set<string>;
+  addLoadedStyle: (style: string) => void;
```
tmeyer2115, line 3: In the future, we will want to support (at least some) HMR for user styles. As such, I think it could be good to have a dedicated slice for styles specifically. We can add them and remove them as users add `.css` imports (or remove them) to their files. Maybe in the future we have a more complex Loading State and we'd want a Slice like this, but for now I think the Loading Progress can be properly derived without it.

packages/studio-ui/src/AppWithLazyLoading.tsx
```diff
@@ -15,24 +15,26 @@ export default function AppWithLazyLoading() {
...
     void AppPromise.then(() => setAppLoaded(true));
   }, []);
 
   return (
     <LoadingOverlay
-      loading={!componentsLoaded || !appLoaded}
+      loading={!componentsLoaded || !resourcesLoaded}
```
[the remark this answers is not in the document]
alextaing: done!

packages/studio-ui/src/store/models/slices/LoadingProgressSlice.ts
alextaing: Spoke offline, renamed to loadedStyles!

tmeyer2115 (approved):

packages/studio-ui/src/AppWithLazyLoading.tsx
```diff
@@ -15,24 +15,26 @@ export default function AppWithLazyLoading() {
...
     loadComponents();
-    loadStyling();
+    void Promise.all(loadStyling()).then(() => setStylesLoaded(true));
     void AppPromise.then(() => setAppLoaded(true));
   }, []);
 
   return (
     <LoadingOverlay
-      loading={!componentsLoaded || !appLoaded}
+      loading={!(componentsLoaded && stylesLoaded && appLoaded)}
```
nmanu1, line 29: nit: can we use `componentLoaded && resourcesLoaded` here?

alextaing: done!

packages/studio-ui/src/utils/loadUserAssets.tsx
```diff
@@ -57,22 +58,24 @@ export function loadStyling(): void {
...
-      document.head.appendChild(styleEl);
-      styleEl.disabled = true;
-    });
+  return userStyleFilepaths.map((filepath) => {
+    return dynamicImportFromBrowser(filepath)
+      .then((styling) => {
+        const styleEl = document.createElement("style");
+        styleEl.innerText = styling.default;
+        styleEl.setAttribute(USER_CUSTOM_STYLE_ATTRIBUTE, filepath);
+        document.head.appendChild(styleEl);
```
nmanu1, line 67: Now that we have a slice in store for the loaded styles, do we still need to add these style tags to the parent document head? Or could we do the dynamic import and add the styling in the slice so `useInjectUserStyles` can just construct the tags from there?

alextaing: yes! We can do that in a future PR

tmeyer2115 (approved):

PR #422, opened 2023-10-25
Add Tailwind to Test Site and E2E Tests

alextaing:
This PR configures Tailwind in our test site and end-to-end tests.

J=SLAP-2912

### DIFF
--- a/e2e-tests/tailwind.config.ts
+++ b/e2e-tests/tailwind.config.ts
@@ -1,0 +1,15 @@
+import type { StudioTailwindConfig } from "@yext/studio";
+
+export default {
+  content: ["./src/**/*.{ts,tsx}"],
+  theme: {
+    extend: {
+      colors: {
+        primary: "#aa00ff",
+      },
+      fontSize: {
+        medium: "14px",
+      },
+    },
+  },
+} satisfies StudioTailwindConfig;
--- a/apps/test-site/package.json
+++ b/apps/test-site/package.json
@@ -25,10 +25,13 @@
     "@yext/pages": "^1.0.0-rc.1",
     "@yext/studio": "*",
     "@yext/types": "^0.1.12-alpha",
+    "autoprefixer": "^10.4.16",
     "eslint": "^8.11.0",
     "eslint-plugin-import": "^2.25.4",
     "eslint-plugin-react-perf": "^3.3.1",
-    "eslint-plugin-tsdoc": "^0.2.17"
+    "eslint-plugin-tsdoc": "^0.2.17",
+    "postcss": "^8.4.31",
+    "tailwindcss": "^3.3.4"
   },
   "scripts": {
     "dev": "studio",
--- a/apps/test-site/postcss.config.cjs
+++ b/apps/test-site/postcss.config.cjs
@@ -1,0 +1,6 @@
+module.exports = {
+  plugins: {
+    tailwindcss: {},
+    autoprefixer: {},
+  },
+};
--- a/apps/test-site/src/layouts/LocationLayout.tsx
+++ b/apps/test-site/src/layouts/LocationLayout.tsx
@@ -1,4 +1,5 @@
 import Header from "../components/Header";
+import "../styles/main.css";
 
 function LocationLayout() {
   return (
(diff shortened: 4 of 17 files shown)


e2e-tests/tailwind.config.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default {
+  content: ["./src/**/*.{ts,tsx}"],
+  theme: {
+    extend: {
+      colors: {
+        primary: "#aa00ff",
+      },
+      fontSize: {
+        medium: "14px",
```
nmanu1, line 11: Do we actually use any of these theme extensions from this config in the tests?

Also, I thought the way we had set up Tailwind currently, is that each test would need to specify the path to the Tailwind config for it to actually be copied over to the cloned playground for that test?

packages/studio-ui/src/hooks/useInjectActiveStyles.tsx
```diff
@@ -10,12 +10,12 @@ import { StudioStore } from "../store/models/StudioStore";
  * styling of the current active page and component tree.  Styling is cleared
  * from the iframe before each injection.
  */
-export default function useInjectUserStyles(
+export default function useInjectActiveStyles(
   iframeDocument: Document | undefined
 ) {
-  const [userStyleImports, loadedStyleFilepaths] = useStudioStore(
+  const [activeStylesFilepaths, loadedStyleFilepaths] = useStudioStore(
```
nmanu1, line 16: nit: It's a little odd to me that one of these variable names has "styles" plural and the other has "style". Can we pick one? I'm slightly leaning toward "\<blah\>StyleFilepaths" so there's not as many plurals

packages/studio-ui/src/hooks/useInjectActiveStyles.tsx
```diff
@@ -25,7 +25,7 @@ export default function useInjectUserStyles(
     if (!iframeDocument) {
       return;
     }
-    userStyleImports.forEach((styleFilepath) => {
+    activeStylesFilepaths.forEach((styleFilepath) => {
       injectStyleIntoIframe(iframeDocument, styleFilepath);
```
nmanu1, line 29: This isn't really a question about this PR, but just to refresh my understanding here, now that we're manually going through dependency trees to get all the CSS filepaths and dynamically importing them, why do we still need the Vite plugin to add the `inline` query param? Is that something we could just do when making the dynamic import?

alextaing: When we dynamically import components, they may also depend CSS. So, Vite will try to add style tags when traversing file dependencies if it comes across a style file, so we want to make sure that we add the inline tag to any style file that Vite touches.  This is also generally the case for transitive dependencies.  If we imported something with the `inline` query param, it won't be propagated when vite imports that file's dependencies so we'd need to add it ourselves.

packages/studio-ui/src/hooks/useInjectActiveStyles.tsx
alextaing: changed to the one with less plurals!

tmeyer2115 (approved):

e2e-tests/tailwind.config.ts
alextaing: I see-- yeah, It looks like we don't use the extended theme.  I also saw that we didn't use it in the test-site either, so I removed it here as well.

And yes!  Thanks for the catch, I see what you're saying.  I added the tailwind to `studioTest` so that the tailwind is correctly copied over into the playground.  I tested with a custom theme extension, and it works!

tmeyer2115 (approved):

e2e-tests/tests/infra/studioTest.ts
```diff
@@ -25,7 +25,7 @@ export type StudioTestFixtures = {
 export const studioTest = base.extend<StudioTestFixtures>({
   createRemote: false,
   debug: false,
-  tailwindConfigPath: undefined,
+  tailwindConfigPath: "tests/infra/tailwind.config.ts",
```
nmanu1, line 28: Now that we have this default Tailwind config for every test, can we rename the config we use in the custom Tailwind test to indicate that it's different and has a custom theme?

alextaing: done!

nmanu1 (approved):

PR #423, opened 2023-10-25
Ensure `_client` and `_server` Templates Don't Appear in Studio.

tmeyer2115:
This PR ensures Studio will not display the `_client.tsx` or `_server.tsx` Templates. These Templates are reserved files in PagesJS. They don't actually correspond to a Page, but instead dictate how SSR works. In the future, we will be able to exclude these files by supporting Globs in the Studio Config's paths. For now, Product does not have the appetite for adding Globs. 

We now prevent someone from adding a Page in Studio whose filename is `_client` or `_server` (assuming they're in PagesJS mode). 

J=SLAP-2960
TEST=auto, manual

### DIFF
--- a/packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
+++ b/packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
@@ -214,7 +214,14 @@ export default class ParsingOrchestrator {
         `The pages directory does not exist, expected directory to be at "${this.paths.pages}".`
       );
     }
-    return createFilenameMapping(this.paths.pages, this.getOrCreatePageFile);
+
+    const excludeReservedPagesJSFiles = (filename: string) =>
+      !(filename.includes("_server") || filename.includes("_client"));
+    return createFilenameMapping(
+      this.paths.pages,
+      this.getOrCreatePageFile,
+      excludeReservedPagesJSFiles
+    );
   }
 
   private getSiteSettings(): SiteSettings | undefined {
--- a/packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
+++ b/packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
@@ -215,13 +215,18 @@ export default class ParsingOrchestrator {
       );
     }
 
-    const excludeReservedPagesJSFiles = (filename: string) =>
-      !(filename.includes("_server") || filename.includes("_client"));
-    return createFilenameMapping(
-      this.paths.pages,
-      this.getOrCreatePageFile,
-      excludeReservedPagesJSFiles
-    );
+    if (this.studioConfig.isPagesJSRepo) {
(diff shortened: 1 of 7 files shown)


packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -214,7 +214,14 @@ export default class ParsingOrchestrator {
         `The pages directory does not exist, expected directory to be at "${this.paths.pages}".`
       );
     }
-    return createFilenameMapping(this.paths.pages, this.getOrCreatePageFile);
+
+    const excludeReservedPagesJSFiles = (filename: string) =>
```
nmanu1, line 218: Do we only want this filter if `isPagesJSRepo` is true?

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
```diff
@@ -214,7 +214,14 @@ export default class ParsingOrchestrator {
...
+      !(filename.includes("_server") || filename.includes("_client"));
```
nmanu1, line 219: I can't remember the exact details of our page name validation in `studio-ui`. If a user tried to add a new page with the name `_server` through the UI, will they currently get an error about it being an invalid filename because it starts with `_`?

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
tmeyer2115: Updated!

packages/studio-plugin/src/orchestrators/ParsingOrchestrator.ts
tmeyer2115: Our validation doesn't seem to catch the `_`. I was able to make a Page Template via Studio  whose page name was `_server`.

nmanu1: I see. Then maybe we should specifically block users from creating a page with either of the reserved names because that could cause issues when saving

tmeyer2115: I briefly mentioned this in my commit message. I was planning to wait for the Glob work to add this kind of validation. With Globs, we will also incorporate support for sub-folders. If a Developer restricts Studio users to certain sub-folders of `src/templates`, we'll want to ensure any Pages they create live in one of the approved sub-folders. So, our validation for Page Creation will need a non-trivial overhaul. As part of that overhaul, we'd check for blacklisted .tsx files in the Glob (such as `_client` and `_server`) and make sure new Pages didn't have those names either. 

I can add some quick validation for now though.

tmeyer2115:
Screenshot 2023-10-25 at 12 00 42 PM
Screenshot 2023-10-25 at 12 00 52 PM

nmanu1 (approved):

PR #424, opened 2023-10-25
Fix viewport selection bug

nmanu1:
This PR fixes a bug with viewport selection where some changes to aspect ratio didn't properly apply the correct style (i.e. applied `w-full` instead of `h-full` or vice versa). Aaron also had some confusion as to why the size of the iFrame didn't match the pixel values displayed for each viewport. So, to fix both things, I updated the viewport styling to use exact pixel values instead of the aspect ratio. Now, if a viewport is wider than the size of the preview panel, the overflow width is hidden but scrollable. But, even our largest viewports can still be viewed in full by collapsing the sidebars. If the viewport is longer than the size of the preview panel, the length of the Studio page expands to encompass it.

Also, this fixes another odd UX case where changing the active component would change the iFrame size because if one component had a lot of props, the prop editing panel on the right side would be very long and increase the length of the page. And since we were using aspect ratio and if it was fit to `h-full`, then the change in page height was cause the "zoom" of the iFrame to increase, which would look like an unexpected width increase in the iFrame despite being on the same viewport when switching between active components.

J=SLAP-2970
TEST=manual

Tested that the bug Aaron reported was no longer occurring locally or in Studio in Storm. Saw that exact pixel values were used for the iFrame width and height instead of aspect ratios.

### DIFF
--- a/packages/studio-ui/src/components/IFramePortal.tsx
+++ b/packages/studio-ui/src/components/IFramePortal.tsx
@@ -1,16 +1,7 @@
 import { createPortal } from "react-dom";
-import {
-  Dispatch,
-  PropsWithChildren,
-  RefObject,
-  SetStateAction,
-  useEffect,
-  useRef,
-  useState,
-} from "react";
+import { Dispatch, PropsWithChildren, SetStateAction, useEffect } from "react";
 import useStudioStore from "../store/useStudioStore";
 import { twMerge } from "tailwind-merge";
-import { Viewport } from "./Viewport/defaults";
 
 export default function IFramePortal(
   props: PropsWithChildren<{
@@ -19,18 +10,13 @@ export default function IFramePortal(
     setIframeEl: Dispatch<SetStateAction<HTMLIFrameElement | null>>;
   }>
 ) {
-  const previewRef = useRef<HTMLDivElement>(null);
   const iframeDocument = props.iframeEl?.contentWindow?.document;
-  const [viewport] = useStudioStore((store) => [store.pagePreview.viewport]);
   useParentDocumentStyles(iframeDocument);
-  const iframeCSS = twMerge(
-    "mr-auto ml-auto",
-    viewport.css,
-    useCSS(viewport, previewRef)
-  );
+  const viewportCss = useStudioStore((store) => store.pagePreview.viewport.css);
+  const iframeCSS = twMerge("mr-auto ml-auto", viewportCss);
 
   return (
-    <div ref={previewRef} className="grow w-1/3 bg-white border-y-8">
+    <div className="grow w-1/3 bg-white border-y-8 overflow-x-scroll">
       <iframe
         id="iframe"
         title={props.title}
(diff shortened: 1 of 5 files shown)


alextaing:
The behavior for super long aspect ratios (e.g. Galaxy Z Flip5 Unfolded) seems kind of funky. Has product confirmed that that's what they want?

packages/studio-ui/src/components/Viewport/defaults.tsx
```diff
@@ -47,7 +47,7 @@ export const VIEWPORTS: ViewportMap = {
       width: 720,
     },
     type: "mobile",
-    css: "aspect-[720/748]",
+    css: "w-[720px] h-[748px]",
   },
   galaxyzflip5unfolded: {
```
alextaing, line 52: The behavior for super long aspect ratios (e.g. Galaxy Z Flip5 Unfolded) seems kind of funky. Has product confirmed that that's what they want?

nmanu1: Out of curiosity, what feels weird about it for you? I still haven't heard back from Aaron, but I'll ping him again about it

alextaing: I think part of it is that since it is a phone screen, as a user I'd expect to be able to see everything since in my mind a phone screen is smaller than a computer screen.

nmanu1: I see. I showed Aaron that specific behavior as well and he's fine with it for now

packages/studio-ui/src/components/IFramePortal.tsx
```diff
@@ -55,32 +41,3 @@ function useParentDocumentStyles(iframeDocument: Document | undefined) {
...
-      (viewport.styles?.width ?? 0) * (previewRef.current?.clientHeight ?? 0) >
-      (viewport.styles?.height ?? 0) * (previewRef.current?.clientWidth ?? 0)
-        ? " w-full "
-        : " h-full ";
-    if (tempCss !== css) {
-      setCss(tempCss);
-    }
-  };
-
-  useEffect(() => {
```
alextaing, line 77: Just curious, do you know what this was doing before? Since it looks like we were already applying CSS to the iframe.  Other than this question, LGTM!

nmanu1: Originally, we were using the aspect ratio instead of the actual pixel sizes to determine the height and width of the viewport. As Alexis [mentioned](https://github.com/yext/studio/pull/336#discussion_r1297305754) in her PR, aspect ratio needs a height or width specified in order to work properly. So, this `useEffect` was in charge of adding either the `w-full` or `h-full` style to the iFrame depending on which viewport dimension we could max out while still fitting within the preview panel size. This hook updated the styling any time the viewport changed or the preview panel's size changed

alextaing: ahh I see.  That makes sense, thanks!

PR #426, opened 2023-10-30
Default to Running Studio Without Strict Mode

alextaing:
This PR adds a flag that will render Studio in React Strict Mode.  Now, Studio runs without React Strict Mode by default, but can be specified to run in Strict Mode by adding `strict` as an argument to the empty `--` flag (i.e. `npx studio -- strict`).  This was also added to the `dev` command in the test site, but not the e2e-tests.

Since we are trying to access an env variable to determine a conditional render in Studio client-side, we cannot access `process.env` directly.  Vite offers `import.meta.env` as a way for client-side modules to access env vars, but they must be prefixed with `VITE_`.  This is why the env var is called `VITE_STUDIO_STRICT`.

J=SLAP-2966
TEST=manual


nmanu1 (commented):
Did we want to look into [this](https://yexttest.atlassian.net/browse/SLAP-2955) approach to achieve similar behavior?

alextaing:
> Did we want to look into [this](https://yexttest.atlassian.net/browse/SLAP-2955) approach to achieve similar behavior?

Talked with Tom, and I think that he wants to get Strict mode working for now, then look into this later.  I think there's a lot to test with regards to that change, and I think we want to get some QA feedback addressed first.  We'll keep the item around, though!

nmanu1 (commented):
Out of curiosity, what's the reasoning behind not wanting strict mode in e2e-tests?

packages/studio/bin/studio.ts
```diff
@@ -32,6 +33,7 @@ cli
         env: {
           ...process.env,
           YEXT_STUDIO_ARGS: JSON.stringify(options),
+          VITE_STUDIO_STRICT: String(useStrictMode),
```
nmanu1, line 36: Are the possible values for `VITE_STUDIO_STRICT` just the strings "true" and "false", or could it also be "undefined" in the case where `options["--"]` above is undefined?

alextaing: Whether or not the `--` flag is specified/has arguments, `options ["--"]`  always returns an array. Either an empty one, or one with values.  So `options ["--"]` should always be defined.  I changed the type of `CliArgs` to reflect that `--` should never be undefined!

nmanu1 (approved):

